PR #46, opened 2023-10-10
feat: integration tests setup

Romsters:
# What ❔

Integration tests setup

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   app:
     build:


docker-compose.yaml
```diff
@@ -1,3 +1,5 @@
+version: '3.8'
```
vasyl-ivanchuk, line 1: @pcheremu do we need a version here? I would rather not add a version if there is no reason for it.

.github/workflows/app-e2e.yml
```diff
@@ -79,8 +79,6 @@ jobs:
         uses: actions/cache@v3
         env:
           cache-name: cache-node-modules
-          # Workaround for bug https://github.com/typicode/husky/issues/991
-          HUSKY: 0
         with:
```
vasyl-ivanchuk, line 82: Just to confirm, we don't need `HUSKY: 0` anymore and tests successfully work without it, do we?

vasyl-ivanchuk (changes requested):

docker-compose.yaml
pcheremu: @vasyl-ivanchuk Yes, it does not work on CI without a version. I have tried again https://github.com/matter-labs/block-explorer/actions/runs/7889736145/attempts/1?pr=46

.github/workflows/app-e2e.yml
pcheremu: I researched the issue in git threads, and have found that this is not actual already. I tried both options (with it and without), for us - no difference.

vasyl-ivanchuk (approved):

PR #50, opened 2023-10-12
chore: separate app config for local and dev

vasyl-ivanchuk:
# What ❔

Separate App configurations for local and dev setups.

## Why ❔

For development setup we're gonna show all networks, but for local (which will be used when running from CLI) only Local network will be shown.

**Dev (default if you just run the application):**
Screenshot 2023-10-12 at 12 23 53

**Local (will be used by CLI):**
Screenshot 2023-10-12 at 12 24 29

## Checklist

- [X] Code has been formatted via `zk fmt` and `zk lint`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -4,8 +4,6 @@ services:
     build:
       context: .
       dockerfile: ./packages/app/Dockerfile
-    environment:
-      - VITE_APP_ENVIRONMENT=local
     command: npm run --prefix packages/app dev -- --host
     ports:
       - '3010:3010'
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -4,9 +4,14 @@ services:
     build:
       context: .
       dockerfile: ./packages/app/Dockerfile
+    environment:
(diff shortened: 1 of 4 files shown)


docker-compose.yaml
```diff
@@ -26,14 +29,14 @@ services:
       - DATABASE_PASSWORD=postgres
       - DATABASE_NAME=block-explorer
       - BLOCKCHAIN_RPC_URL=http://zksync:3050
+      - BATCHES_PROCESSING_POLLING_INTERVAL=1000
```
vasyl-ivanchuk, line 32: Decrease the polling interval for batches on local env. Currently the problem is that on the first BatchesProcessor run L1 batches are not yet ready to be processed, so they will be processed only in 60 seconds on the next BatchesProcessor run. As a result, a user won't see batches information for a while when they explore the BE UI. With the decreased timeout batches and all related info appear quickly. I don't think that the frequency of the polling will be an issue on local.

Romsters (approved):

PR #52, opened 2023-10-13
chore: docker compose for cli

vasyl-ivanchuk:
# What ❔

Docker compose file for CLI. Multi stage Docker file for App package.

## Why ❔

We need a slightly different docker compose configuration for CLI, so it was added.
Docker file for App was changed to multi stage to make build consistent across all packages and introduce different stages for dev and prod envs.

## Checklist

### DIFF
--- a/docker-compose-cli.yaml
+++ b/docker-compose-cli.yaml
@@ -5,7 +5,6 @@ services:
     build:
       context: .
       dockerfile: ./packages/app/Dockerfile
-    command: npm run --prefix packages/app dev -- --host
     environment:
       - VITE_APP_ENVIRONMENT=local
     ports:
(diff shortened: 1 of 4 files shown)


docker-compose-cli.yaml
```diff
@@ -0,0 +1,75 @@
+version: '3.2'
+
+services:
+  app:
+    build:
+      context: .
+      dockerfile: ./packages/app/Dockerfile
+    command: npm run --prefix packages/app dev -- --host
```
Romsters, line 8: why dev? For cli I'd suggest build + start

vasyl-ivanchuk: Fixed

Romsters (dismissed):

Romsters (approved):

PR #53, opened 2023-10-13
fix: do not request verification info when running local node

JackHamer09:
# What ❔
- Do not request contract verification info when running against local node

## Why ❔
- Since contract verification is not yet implemented in the Block Explorer API the request simply fails and any contract page becomes unavailable


vasyl-ivanchuk (changes requested):
1. `verificationApiUrl` should also be removed from `/packages/app/src/configs/dev.config.json` for local network.
2. we have tests that check `verificationInfo` in both `useAddress` and `useContractAPI`, please add a case to check that `null` value is returned if there is no `verificationApiUrl` configured.

vasyl-ivanchuk (approved):

PR #59, opened 2023-10-18
fix: return eth token even if it does not exist in db

vasyl-ivanchuk:
# What ❔

API changes to return ETH token even if it does not exist in DB yet.

## Why ❔

ETH transfers might appear before we get information about system contract deployment so even if ETH is not in DB we should return it from API endpoints for ETH token and ETH transfers endpoints.

## Checklist


docker-compose-cli.yaml
```diff
@@ -5,8 +5,6 @@ services:
     build:
       context: .
       dockerfile: ./packages/app/Dockerfile
-    environment:
-      - VITE_APP_ENVIRONMENT=local
```
vasyl-ivanchuk, line 9: Not needed anymore. We need to setup `appEnvironment` with runtime and not with build config, since we will use the same built version for different app environments.

docker-compose-cli.yaml
```diff
@@ -58,8 +49,6 @@ services:
       driver: none 
     volumes:
       - postgres:/var/lib/postgresql/data
-    ports:
-      - "5432:5432"
```
vasyl-ivanchuk, line 62: Hide all exposed ports except BE UI (`3010`) to reduce the number of possible conflicts with other running apps and containers.

packages/api/src/transfer/transfer.entity.ts
```diff
@@ -53,9 +53,9 @@ export class Transfer extends BaseEntity {
   @Column({ type: "varchar", length: 128, nullable: true })
   public readonly amount?: string;
 
-  @ManyToOne(() => Token)
+  @ManyToOne(() => Token, { createForeignKeyConstraints: false })
```
vasyl-ivanchuk, line 56: By default ManyToOne creates a foreign key, but in worker schema this entity doesn't have it:
1. it creates a discrepancy between e2e DB and DB created by worker.
2. requires creating some additional entities in e2e tests even if they are not used.

To solve this issue createForeignKeyConstraints relation option is used so DB created by API for e2e and DB created by worker match.

Romsters (approved):

PR #69, opened 2023-10-26
feat: public endpoint to verify contracts

vasyl-ivanchuk:
# What ❔

Public endpoint to verify contracts

## Checklist

- [X] Documentation comments have been added / updated.


packages/api/package.json
```diff
@@ -62,12 +62,12 @@
     "@typescript-eslint/parser": "^5.0.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
-    "jest": "28.1.3",
+    "jest": "29.2.1",
```
vasyl-ivanchuk, line 65: Update `jest` to fix an issue with open handlers for supertest request post functions.

packages/api/src/api/api.controller.ts
```diff
@@ -38,11 +48,11 @@ export class ApiController {
   @ApiExcludeEndpoint()
   @Get("api")
   @UseFilters(ApiExceptionFilter)
-  public async apiHandler(
+  public async apiGetHandler(
     @Req() request: Request,
     @Next() next: NextFunction,
-    @Query("module", new ParseModulePipe()) module: ApiModule,
     @Query(new ParseActionPipe()) action: string,
+    @Query("module", new ParseModulePipe()) module: ApiModule,
```
vasyl-ivanchuk, line 55: Change the order so module pipe works before action pipe. Otherwise even if the module is not found, error about action will be returned.

packages/api/src/configureApp.ts
```diff
@@ -3,7 +3,7 @@ import { INestApplication, ValidationPipe } from "@nestjs/common";
 export const configureApp = (app: INestApplication) => {
   app.useGlobalPipes(
     new ValidationPipe({
-      disableErrorMessages: true,
+      disableErrorMessages: false,
```
vasyl-ivanchuk, line 6: To return the description of 400 Bad Request Errors

packages/worker/package.json
```diff
@@ -61,13 +61,13 @@
     "@typescript-eslint/parser": "^5.0.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
-    "jest": "28.1.3",
+    "jest": "29.2.1",
```
vasyl-ivanchuk, line 64: Same update as for API, just to use the lib versions.

packages/worker/src/app.service.spec.ts
```diff
@@ -215,10 +215,10 @@ describe("AppService", () => {
...
       expect(blocksRevertService.handleRevert).toBeCalledWith(blockNumber);
 
-      expect(blockService.start).toBeCalledTimes(1);
-      expect(batchService.start).toBeCalledTimes(1);
-      expect(counterService.start).toBeCalledTimes(1);
-      expect(balancesCleanerService.start).toBeCalledTimes(1);
+      expect(blockService.start).toBeCalledTimes(2);
+      expect(batchService.start).toBeCalledTimes(2);
+      expect(counterService.start).toBeCalledTimes(2);
+      expect(balancesCleanerService.start).toBeCalledTimes(2);
```
vasyl-ivanchuk, line 221: Honestly, not sure how it worked before but these checks started to fail after`jest` update. 2 - is the correct number of calls here, since 1st call happens on app init, and then the 2nd - while processing the revert event.

Romsters (approved):

PR #91, opened 2023-11-17
feat: use new tokens api

Romsters:
# What ❔

- use new tokens API
- remove token library
- remove token price API usage
- refactoring


vasyl-ivanchuk (dismissed):

JackHamer09:
I see that there are only 42 tokens in the list on mainnet. It seems because of the liquidity filter. Is this Era liquidity for these tokens? Cause the list looks really weird tbh.
Also what about the testnets? Do we have any token data for testnets?

Romsters:
> I see that there are only 42 tokens in the list on mainnet. It seems because of the liquidity filter. Is this Era liquidity for these tokens? Cause the list looks really weird tbh. Also what about the testnets? Do we have any token data for testnets?

There are more tokens displayed now in fact, and yes, some of the tokens are not displayed because of the liquidity filter. This is not Era liquidity, this is L1 liquidity. We don't have tokens data for testnet, I think this is fine, etherscan also doesn't have testnet tokens data as there is no liquidity on testnet.

PR #109, opened 2023-12-04
feat: return gasUsed, gasPerPubdata, maxFeePerGas and maxPriorityFeePerGas fields for transaction

vasyl-ivanchuk:
# What ❔

Return `gasUsed`, `gasPerPubdata`, `maxFeePerGas` and `maxPriorityFeePerGas` fields for transaction.

## Why ❔

For better UX we want to show gas and fee related fields on UI so we need to return these fields from the API first.

## Checklist


packages/api/src/transaction/entities/transactionDetail.entity.ts
```diff
@@ -0,0 +1,18 @@
+import { Entity } from "typeorm";
+import { Transaction } from "./transaction.entity";
+
+@Entity({ name: "transactions" })
+export class TransactionDetail extends Transaction {
```
Romsters, line 5: TransactionDetail -> TransactionDetails?

Romsters (approved):

PR #141, opened 2023-12-22
Automatic system alerts

Romsters:


packages/app/src/components/IndexerDelayAlert.vue
```diff
@@ -0,0 +1,77 @@
...
+  const latestBlockDate = new Date(latestBlock.value.timestamp * 1000);
+  const latestIndexedBlockDate = new Date(latestIndexedBlocks.value[0].timestamp);
+  const delay = latestBlockDate.getTime() - latestIndexedBlockDate.getTime();
+  return delay;
+});
+
+const indexerDelayInHours = computed(() => (indexerDelay.value / (1000 * 60 * 60)).toFixed(1));
+
+const isIndexerDelayed = computed(() => {
+  if (!latestBlock.value?.number || !latestIndexedBlocks.value?.length) {
```
vasyl-ivanchuk, line 51: This check is not needed, since it is already handled in `indexerDelay` computed. It seems that `return indexerDelay.value > MIN_DELAY_TO_SHOW_ALERT;` is enough.

packages/app/src/components/common/SystemAlert.vue
vasyl-ivanchuk, line 1: CSS for links has to added, otherwise they are of the similar color as background:
image
You can check how links look like on the Goerli deprecation message.

packages/app/src/components/header/TheHeader.vue
```diff
@@ -305,10 +305,6 @@ const hasContent = computed(() => {
   .hero-banner-container {
     @apply absolute left-0 top-full flex h-64 w-full items-end justify-end overflow-hidden bg-primary-900;
 
-    &.mainnet {
```
vasyl-ivanchuk, line 308: This hack was needed to increase the height of the header background (which has a fixed size) when a warning message is shown. Otherwise the header text is pushed down and is not visible:
image
Having dynamic error messages we will also need to change the height of the background image dynamically or rework the whole thing.

packages/app/src/locales/en.json
```diff
@@ -721,5 +721,9 @@
                 "callData": "calldata"
             }
         }
+    },
+    "systemAlert": {
+        "indexerDelayed": "Transaction indexing is {indexerDelayInHours} hours behind. Transactions are being processed normally and will gradually show up. You can also use other explorers like <a href=\"https://hyperscan.xyz/\">hyperscan.xyz</a> meanwhile.",
```
vasyl-ivanchuk, line 726: I'm concerned about the wording here. Users may misunderstood that _Indexing is N hours behind_ means that only in N hours they will see the data they have just submitted. But in fact, it's a delay at the moment, but the Explorer may catch up much quicker.

Not sure what is the better option, I'm just thinking out loud.

vasyl-ivanchuk (changes requested):

packages/app/src/locales/en.json
Romsters: Agree, but I think it's more or less good to start with. I think showing number of blocks behind might be confusing.

packages/app/src/components/header/TheHeader.vue
Romsters: We generally don't support mobile, but this should be fixed properly. Out of scope of this PR I'd say.

packages/app/src/components/IndexerDelayAlert.vue
Romsters: agree

Romsters: removed

vasyl-ivanchuk (approved):

PR #149, opened 2024-01-11
chore: retry wrapper over api requests

abilevych:
# What ❔

The API retry logic for integration tests

## Why ❔

It allows increase speed of test runs due to refusing the timeouts-based approach

## Checklist

### DIFF
--- a/packages/integration-tests/src/helper.ts
+++ b/packages/integration-tests/src/helper.ts
@@ -82,7 +82,7 @@ export class Helper {
         if (i === localConfig.maxAPIretries - 1) {
           throw error;
         }
-        await new Promise((resolve) => setTimeout(resolve, localConfig.intervalAPIretries));
+        await this.delay(localConfig.intervalAPIretries);
       }
     }
   }


packages/integration-tests/src/helper.ts
```diff
@@ -55,4 +56,34 @@ export class Helper {
...
+  async retryTestAction(action) {
+    for (let i = 0; i < localConfig.maxAPIretries; i++) {
+      try {
+        await action();
+        return;
+      } catch (error) {
+        if (i === localConfig.maxAPIretries - 1) {
+          throw error;
+        }
+        await new Promise((resolve) => setTimeout(resolve, localConfig.intervalAPIretries));
```
pcheremu, line 85: Not sure, but may be here you planned to use: await delay(localConfig.intervalAPIretries)

pcheremu: The solution looks elegant. Im just surprising, could we make a DDoS behavior using 200 iteration for each action() ? 
Probably you tried it, Im just interesting will we have the same effect or less DDoS in case of: 
```
maxAPIretries: 200, ---> 40
intervalAPIretries: 0.2 * 1000, ---> 1 * 1000
```
 or in some other similar proportion compensating iterations by increasing the timeout?

pcheremu (commented):
I'm going to make the review one more time tomorrow morning with fresh feelings. For now tests looks good and the logic is elegant. No idea if we need to worry about it or 200 iterations with low latency is preferable than longer latency with less amount of iterations. I have also noticed that (despite on some excluded tests) the total time for api tests was decreased ~ on 23% (from 17min to 13min).

pcheremu (approved):

PR #150, opened 2024-01-16
fix: docs links

vasyl-ivanchuk:
# What ❔

Fix links to documentation.

## Why ❔

The documentation has been updated so we should use new links.

## Checklist


MexicanAce (approved):
LGTM

PR #159, opened 2024-01-30
fix: remove portal from header

vasyl-ivanchuk:
# What ❔

- Remove Portal link from the top menu
- Do not show an error reason for the failed transaction if it's empty
- Docker compose starts services in production and not in development mode
- Add zkVM filter option for solc compilers on the verification page

## Why ❔

To fix bug reports / implement feature requests.

## Checklist


packages/app/src/components/form/FormItem.vue
```diff
@@ -42,7 +42,7 @@ defineProps({
 }
 .label-inline-block {
   .form-item-label {
-    @apply inline-block;
+    @apply float-left;
```
Romsters, line 45: Are you sure it won't break any existing use cases?

packages/app/src/components/common/CheckBoxInput.vue
```diff
@@ -0,0 +1,46 @@
...
+export default {
+  inheritAttrs: false,
+};
+</script>
+
+<script setup lang="ts">
+import { computed } from "vue";
+const props = defineProps({
+  modelValue: {
+    type: [Boolean],
```
Romsters, line 18: Why not just `type: Boolean` ?

vasyl-ivanchuk: Good point, fixed.

packages/app/src/components/form/FormItem.vue
vasyl-ivanchuk: I don't think so. This class used in two places and I checked them both.

Romsters (approved):

PR #170, opened 2024-02-14
feat: new automation scope for tests on Contracts API + Tokens API

olehbairak:
# What ❔

Implemented automation test coverage for next test cases:
https://linear.app/matterlabs/issue/QA-653/%5Btokensapi-contracts-api%5D-automate-test-cases-for-contracts-api-for-be

## Why ❔

To check this functionality in PR automation integration tests.

## Checklist


packages/integration-tests/src/helper.ts
```diff
@@ -61,8 +61,12 @@ export class Helper {
     return new Promise((resolve) => setTimeout(resolve, ms));
   }
 
-  async performGETrequest(apiRoute: string) {
-    return request(environment.blockExplorerAPI).get(apiRoute);
+  async performGETrequest(apiRoute: string, network = "local") {
```
olehbairak, line 64: This update of the code is designed to be able to call Sepolia API.
Why?
Because for Contracts API we need to have contracts deployed and **verified**.
Why Sepolia?
Because we don't have verification flow implemented for local environments. As a result, we can't implement it. I tried and appeared that it is not supported at all for now.
So?
I decided to use Sepolia contract, that was deployed by me, and won't be changed.
Why such implementation?
Because by default I left network = local, so that we don't need to update all other usages of this function.

packages/integration-tests/tests/api/contracts.test.ts
```diff
@@ -22,6 +22,61 @@ describe("API module: Contract", () => {
...
+        expect(response.body).toStrictEqual(expect.objectContaining({ status: "1" }));
+        expect(response.body).toStrictEqual(expect.objectContaining({ message: "OK" }));
+        expect(typeof response.body.result).toStrictEqual("string");
+      });
+    });
+
+    //id1802
+    it("Verify /api?module=contract&action=getsourcecode response", async () => {
+      await helper.retryTestAction(async () => {
+        const greeterContract = "0x180faDac8ce638C67640a6C29BdfB5CB60eCDB76";
```
olehbairak, line 56: The contract address usage is explained above.

olehbairak:
@pcheremu can you please review the PR?
Thank you in advance.

packages/integration-tests/tests/api/tokens.test.ts
```diff
@@ -113,6 +113,26 @@ describe("Tokens", () => {
         });
       });
 
+      //id1907
+      it("Verify /api?module=token&action=tokeninfo response", async () => {
+        await helper.retryTestAction(async () => {
+          apiRoute = `/api?module=token&action=tokeninfo&contractaddress=${Token.ETHER_ERC20_Address}`;
+          response = await helper.performGETrequest(apiRoute);
```
pcheremu, line 120: It's make sense to align it with all functions and send correct network instead.

packages/integration-tests/src/helper.ts
pcheremu, line 64: Could we make the 2nd argument is optional or is local - the default value for all other cases?

packages/integration-tests/src/helper.ts
```diff
@@ -61,8 +61,12 @@ export class Helper {
...
+    if (network === `local`) {
+      return request(environment.blockExplorerAPI).get(apiRoute);
+    } else if (network === `sepolia`) {
+      return request(environment.blockExplorerSepoliaAPI).get(apiRoute);
+    }
```
pcheremu, line 69:
```suggestion
    } else {
      throw new Error(`The API route for the network ${network} is undefined.`);
    }
```

packages/integration-tests/tests/api/contracts.test.ts
```diff
@@ -22,6 +22,61 @@ describe("API module: Contract", () => {
...
+        expect(response.body).toStrictEqual(expect.objectContaining({ status: "1" }));
+        expect(response.body).toStrictEqual(expect.objectContaining({ message: "OK" }));
+        expect(response.body).toStrictEqual(expect.objectContaining({ result: "Pass - Verified" }));
+      });
+    });
+
+    //id1695
+    it("Verify /api?module=contract&action=getabi response", async () => {
+      await helper.retryTestAction(async () => {
+        const greeterContract = "0x180faDac8ce638C67640a6C29BdfB5CB60eCDB76";
```
pcheremu, line 42: please move it to entities:

export enum Contracts {
  greeterContract = "0x180faDac8ce638C67640a6C29BdfB5CB60eCDB76",
}

pcheremu (changes requested):
@olehbairak please, apply changes based on comments

olehbairak:
> @olehbairak please, apply changes based on comments

@pcheremu thank you for the feedback! 
Already fixed!

pcheremu (commented):
@abilevych all essential changes were applied. I still stacking with helper.performGETrequest(apiRoute). From the one hand we can have an agreement that if we use just one variable it will mean that the network will be local, from the other hand, as for me, it will be more clear to have two arguments helper.performGETrequest(apiRoute, network) to transfer there more explicitly. Both options are make sense, please let us know what would you prefer in this case?

olehbairak:
@pcheremu @abilevych I fixed comments.
Can you please review the changes?
Thank you in advance!

packages/integration-tests/src/helper.ts
olehbairak: @abilevych do you see this comment?

olehbairak:
@abilevych conflicts were resolved.
Now branch is up to date and ready for review.

olehbairak:
> @olehbairak please, apply changes based on comments

@pcheremu can you please resolve the request change - it's currently blocking from merge.
Seems like all the requested changes were made and the PR can be merged.
Thank you in advance!

PR #177, opened 2024-02-15
feat: integration tests setup

pcheremu:
# What ❔
no changes with integration-tests branch except fixed commit signatures
 
Integration tests setup

## Checklist


vasyl-ivanchuk (dismissed):

pcheremu:
@Romsters @vasyl-ivanchuk @abilevych  After the warning "branch is out-of-date", it requires the rebase from the very beginning, with a lot of attention. This PR is too complicated and too heavy to do it regularly. Should I create a simple PR with 1 commit that will be based on the main branch that will contains only changes that belongs to integration-tests?

.github/workflows/integration-tests-api.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: npm
+
+      - name: Install dependencies
+        run: |
+          npm ci --no-audit
```
Romsters, line 40: Why --no-audit?

packages/integration-tests/playwright.config.ts
```diff
@@ -0,0 +1,21 @@
+import { defineConfig, devices } from "@playwright/test";
+import { config } from "tests/ui/config";
+
+export default defineConfig({
+  testDir: "tests/ui",
+  fullyParallel: true,
+  projects: [
+    {
+      name: "chromium",
```
Romsters, line 9: Why the project is called "chromium"? Should dit be just "integration-tests" or something?

.github/workflows/integration-tests-api.yml
```diff
@@ -0,0 +1,61 @@
...
+          npx playwright install --with-deps chromium
```
Romsters, line 41: Why is `playwright install` here? Should we move it to the installation script, maybe postinstall? Otherwise locally I'll have to do the same.

package.json
```diff
@@ -16,6 +16,8 @@
     "test": "lerna run test",
     "test:ci": "lerna run test:ci",
     "test:e2e": "lerna run test:e2e",
+    "test:integration:ui": "lerna run integration-test:ui",
```
Romsters, line 19:
```suggestion
    "test:integration:ui": "lerna run integration-tests:ui",
```

package.json
```diff
@@ -16,6 +16,8 @@
...
+    "test:integration:api": "lerna run integration-test:api",
```
Romsters, line 20:
```suggestion
    "test:integration:api": "lerna run integration-tests:api",
```

packages/integration-tests/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  extends: ["@matterlabs/eslint-config-vue"],
```
Romsters, line 2: Why `eslint-config-vue`? We don't use vue in integration-tests

packages/integration-tests/README.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Setting up env variables
+
+- Create `.env` file in the `integration-tests/src/playbook/` package folder and copy paste `.env.example` content in there.
+```
+cp .env.example packages/integration-tests/src/playbook/.env
+```
+
+- Add your private wallet key to `WALLET_PRIVATE_KEY=` By default we use `richWalletPrivateKey` private key in `integration-tests/src/playbook/utils/getWallet.ts` from `integration-tests/src/entities.ts`.
```
Romsters, line 18:
```suggestion
- Add your private wallet key to `WALLET_PRIVATE_KEY={YOUR_PRIVATE_KEY}`. By default we use `richWalletPrivateKey` private key in `integration-tests/src/playbook/utils/getWallet.ts` from `integration-tests/src/entities.ts`. 
```

packages/integration-tests/README.md
```diff
@@ -0,0 +1,59 @@
...
+all tests:
+
+```bash
+npm run integration-test:ui 
+```
+
+If you need to run the exact test or/and suite you can change 
+key-words from `it` to `fit` (for the test) and `describe` to `fdescribe` for suite.
+
+If you need to exclude some specific test/suites, you can change keywords `it` to `xit` and/or
```
Romsters, line 45:
```suggestion
If you need to exclude some specific test/suite, you can change keywords `it` to `xit` and/or
```

packages/integration-tests/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "integration-tests",
+  "version": "0.0.0",
+  "title": "Integration tests",
+  "description": "The test solution for transaction processing`",
```
Romsters, line 5: Is it just for transaction processing? Should it be `The test solution for Block Explorer API and UI`?

packages/integration-tests/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "jest": "^29.6.2",
+    "lint-staged": "^13.1.0",
+    "prettier": "^2.8.2",
+    "supertest": "^6.3.3",
+    "ts-jest": "^29.0.5",
+    "ts-node": "^10.9.1",
+    "zksync-web3": "^0.14.3"
+  },
+  "dependencies": {
+    "ts-jest-resolver": "^2.0.0"
```
Romsters, line 42: Why is this not in devDependencies?

packages/integration-tests/src/config.ts
```diff
@@ -0,0 +1,20 @@
+import { Wallets } from "./entities";
+
+export const localConfig = {
+  gasLimit: { gasLimit: 10000000 },
```
Romsters, line 4: What is this? Is it for L1? Should it be `l1GasLimit: 10000000` then?

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
+export enum Buffer {
```
Romsters, line 1: What is Buffer? Can we use some dynamic path instead based on contract name, test name or whatever it is related to?

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  txEthWithdraw = "./buffer/txEthWithdraw.txt",
+  txEthWithdrawOtherAddress = "./buffer/txEthWithdrawOtherAddress.txt",
+  txERC20Transfer = "./buffer/txERC20Transfer.txt",
+  emptyWalletPrivateKey = "./buffer/emptyWalletPrivateKey.txt",
+  emptyWalletAddress = "./buffer/emptyWalletAddress.txt",
+  failedState = "./buffer/failedState.txt",
+  customToken = "./buffer/customToken.txt",
+}
+
+export enum Logger {
```
Romsters, line 35: Why is it called logger? I would split it into `TransferType` enum containing "DEPOSIT", "WITHDRAW" and "TRANSFER". And the rest would just move to constants.

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  CUST_Address = "0x0928008B245A76E105E02C522b5d309c0887ecA5",
+  customL2TokenName = "L2 ERC20 token",
+  customL2TokenSymbol = "L2",
+  customL2TokenDecimals = 18,
+  ETHER_PULL_Address = "0x0000000000000000000000000000000000008001",
+  ETHER_Address = "0x0000000000000000000000000000000000000000",
+  ETHER_ERC20_Address = "0x000000000000000000000000000000000000800A",
+}
+
+export enum TransactionsType {
```
Romsters, line 54: This should be named `TransferType`

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  fee = "fee",
+  transfer = "transfer",
+  refund = "refund",
+  withdrawal = "withdrawal",
+}
+
+export enum TransactionsStatus {
```
Romsters, line 61:
```suggestion
export enum TransactionStatus {
```

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  failed = "failed",
+}
+
+export enum Wallets {
+  mainWalletAddress = "0x586607935E1462ab762F438E0A7b2968A4158975",
+  secondWalletAddress = "0x26A4c5Dfe2cA3c9E7E8C417B689F41b6b5745C37",
```
Romsters, line 67:
```suggestion
  secondaryWalletAddress = "0x26A4c5Dfe2cA3c9E7E8C417B689F41b6b5745C37",
```

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  richWalletAddress = "0x36615Cf349d7F6344891B1e7CA7C72883F5dc049",
+  mainWalletPrivateKey = "0x06ac1584dd1cf69f97a784b2b7812cd0c65a867ec997add028cdf56483c1c299",
+  secondWalletPrivateKey = "e14e6e0b3b610411cf15c3a5aa3252cac9e0a40a9bbe67ceb3b5d506f56576fd",
```
Romsters, line 70:
```suggestion
  secondaryWalletPrivateKey = "e14e6e0b3b610411cf15c3a5aa3252cac9e0a40a9bbe67ceb3b5d506f56576fd",
```

packages/integration-tests/src/entities.ts
Romsters, line 1: Why is this file called entities? We should create a config file and constants file and move things there.

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  deposit = "DEPOSIT",
+  withdraw = "WITHDRAW",
+  transfer = "TRANSFER",
+  txHashStartsWith = "0x",
+  textSeparator = "======================= ",
+  txFailedState = "FAILED STATE",
+}
+
+export enum Token {
+  CUST_Address = "0x0928008B245A76E105E02C522b5d309c0887ecA5",
```
Romsters, line 45: What is CUST_Address?

packages/integration-tests/src/entities.ts
```diff
@@ -0,0 +1,77 @@
...
+  customL2TokenName = "L2 ERC20 token",
+  customL2TokenSymbol = "L2",
+  customL2TokenDecimals = 18,
+  ETHER_PULL_Address = "0x0000000000000000000000000000000000008001",
```
Romsters, line 49: Please use the same naming convention within a single enum/object.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+import * as request from "supertest";
+import { Provider } from "zksync-web3";
+
+import { environment, localConfig } from "./config";
+import { Logger } from "./entities";
+
+import type { BaseProvider } from "@ethersproject/providers/src.ts/base-provider";
+
+export class Helper {
+  async txHashLogger(txType: string, txValue: string, tokenName?: string) {
```
Romsters, line 14: This is a function, please rename it to `logTransaction` or something.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (tokenName === undefined) {
+      return console.log(logMessage);
+    } else {
+      return console.log(logMessage, ` ${tokenName}`);
+    }
+  }
+
+  async executeScript(script: string) {
+    const output = execSync(script, { encoding: "utf-8" });
```
Romsters, line 25: Please use `exec` and wrap it with Promise, `execSync` is blocking.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    try {
+      console.log(`> Run NPM Script "${script}":\n`, output);
+      return output;
+    } catch (e) {
+      console.log(e);
+    }
+  }
+
+  async getStringFromFile(fileName: string) {
```
Romsters, line 35: The name is confusing, it should just be `readFile`. I think it's `filePath` not `fileName` in the args. What's the point of adding this `..`? You can just include it in the place where you call it. Or otherwise just pass a filename without path and call it like `readFileFrom{FOLDER_NAME}`.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+    const absoluteRoute = path.join(__dirname, "..", fileName);
+
+    try {
+      return await fs.readFile(absoluteRoute, { encoding: "utf-8" });
+    } catch {
+      console.log(`There is no the expected file: ${fileName}`);
```
Romsters, line 41: Do you want it to return undefined in case there is no file? It doesn't look like, better just let it fail as I assume you always expect the file to exist, so if it doesn't let it fail.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  async getBalanceETH(walletAddress: string, layer: string) {
+    let network: string;
+    let provider: BaseProvider;
+    if (layer == "L1") {
+      network = localConfig.L1Network;
+      provider = ethers.getDefaultProvider(network);
+    } else if (layer == "L2") {
+      network = localConfig.L2Network;
+      provider = new Provider(network);
```
Romsters, line 53: Why initialising new Provider every time you want to get ETH value? Add a module that returns you cached provider.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+      console.log(`Wrong layer: ${layer}`);
+    }
+    return ethers.utils.formatUnits(await provider.getBalance(walletAddress), "wei");
+  }
+
+  async delay(ms: number) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+  }
+
+  async performGETrequest(apiRoute: string) {
```
Romsters, line 64:
```suggestion
  async performGetRequest(apiRoute: string) {
```

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+    } else if (layer == "L2") {
+      network = localConfig.L2Network;
+      provider = new Provider(network);
+    } else {
+      console.log(`Wrong layer: ${layer}`);
+    }
+    return ethers.utils.formatUnits(await provider.getBalance(walletAddress), "wei");
+  }
+
+  async delay(ms: number) {
```
Romsters, line 60: Use `import { setTimeout } from "timers/promises";` instead of that.

packages/integration-tests/src/helper.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * A retry wrapper method to enhance test stability in API testing.
+   * Useful when API response fields may not immediately reflect the expected state,
+   * but can update to the correct response after a delay.
+   * Attempts to execute the action a specified number of times (defined in localConfig.maxAPIretries)
+   * with a delay between attempts (localConfig.intervalAPIretries).
+   * Throws an error if the action consistently fails after all retries.
+   */
+  async retryTestAction(action) {
```
Romsters, line 76:
```suggestion
  async runRetriableTestAction(action) {
```

packages/integration-tests/src/playbook/contracts/L1.sol
```diff
@@ -0,0 +1,12 @@
+//SPDX-License-Identifier: Unlicense
```
Romsters, line 1: It should be called just `TestERC20` or something like that, and why do you use two separate contracts for L1 and L2, it should be the same one, just deploy it with different args if you want.

packages/integration-tests/src/playbook/contracts/Middle.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+interface IGreeter {
+    function greet() external view returns (string memory);
+    function setGreeting(string memory) external;
+}
+
+contract Middle {
```
Romsters, line 9: Why is the file and contract name Middle?

packages/integration-tests/src/playbook/contracts/Root.sol
```diff
@@ -0,0 +1,22 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+contract Root {
```
Romsters, line 4: What is Root?

packages/integration-tests/tests/api/addresses.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  let response;
+
+  describe("/address/{address}", () => {
+    beforeAll(async () => {
+      await playbook.deployNFTtoL1();
+      await playbook.deployNFTtoL2();
+      await playbook.deployMultiCallContracts();
+      await playbook.deployMultiTransferETH();
+      await playbook.deployGreeterToL2();
+      await playbook.useMultiCallContracts();
```
Romsters, line 25: Why using contracts is here? I think what we should do is use ethers and execute testing logic within the tests instead of making a bunch of calls and then testing what happened.
The flow should be:
1) Deploy contracts in `beforeAll`
2) In every test interact with a contract and make proper checks. Interactions can be done without using hardhat scripts, we should just use SDK.

packages/integration-tests/src/entities.ts
pcheremu: removed as unused

packages/integration-tests/src/entities.ts
pcheremu: updated

packages/integration-tests/src/helper.ts
pcheremu: updated

packages/integration-tests/src/helper.ts
pcheremu: updated

packages/integration-tests/src/helper.ts
pcheremu: updated, changed all variables and replaced the path in constants.ts

packages/integration-tests/src/helper.ts
pcheremu: updated

packages/integration-tests/src/helper.ts
pcheremu: added cached providers

packages/integration-tests/src/helper.ts
pcheremu: added

packages/integration-tests/src/playbook/contracts/Middle.sol
pcheremu: This is just the name of the contract that we use in the test suite. The suite contains 3 contracts Root - Middle - Caller

packages/integration-tests/src/playbook/contracts/Root.sol
pcheremu: This is just the name of the contract that we use in the test suite. Its name Root because it must be deployed first and called as a 3rd contract in chain Root <- Middle <- Caller

packages/integration-tests/package.json
pcheremu: updated

packages/integration-tests/src/config.ts
pcheremu: updated

.github/workflows/integration-tests-api.yml
pcheremu: fixed

packages/integration-tests/playwright.config.ts
pcheremu: updated

.github/workflows/integration-tests-api.yml
pcheremu: updated

packages/integration-tests/src/entities.ts
pcheremu: updated

packages/integration-tests/src/entities.ts
pcheremu: renamed

packages/integration-tests/.eslintrc.js
pcheremu: @Romsters what you can advice to use here?

packages/integration-tests/package.json
pcheremu: cannot complete on my local  virtual env: `npm i ts-jest-resolver -D`  npm -i hangs.

packages/integration-tests/src/entities.ts
pcheremu: As far as I know this is the initial way of implementation. I have verified, such changes could be possible but requires a lot of (time) to change almost all dependent files. We could prioritize it for later as a separate task or keep it as is (if it is not an issue). what do you think? @Romsters @abilevych

packages/integration-tests/src/entities.ts
pcheremu: @abilevych, I believe we have such name on purpose to differentiate names between `TransactionsType` data , is not it? Should I rename it?

PR #207, opened 2024-03-15
fix: update goerli deprecation message

vasyl-ivanchuk:
# What ❔

New goerli deprecation message with a specific deprecation date:
image

## Why ❔

To notify users when the deprecation is gonna happen.

## Checklist


MexicanAce (approved):
LGTM

PR #272, opened 2024-07-31
fix: update uptime url

JackHamer09:
# What ❔

Update uptime url

## Why ❔

Previous url is outdated and was taken over for security report bug bounty


uF4No (approved):
🫡

PR #275, opened 2024-08-21
fix: add base token address to hyperchain config

sanekmelnikov:
# What ❔

Adding missing `baseTokenAddress` to the hyperchain configuration.
Adding a fallback to use the default L2 base token address if it's not specified in the config.

## Why ❔

When configuring the block explorer using `npm run hyperchain:configure`, the resulting configuration was missing the `baseTokenAddress` value. This absence was causing an error [here](https://github.com/matter-labs/block-explorer/blob/main/packages/app/src/composables/useEnvironmentConfig.ts#L31). As a result of this error, the system was defaulting to Sepolia (which I assume is not the intended behavior for your hyperchain setup).

## Checklist


vasyl-ivanchuk:
These changes are merged in [this](https://github.com/matter-labs/block-explorer/pull/276) PR.

PR #286, opened 2024-09-13
fix for default max gas on payable functions

CryptoMan0x:
# What ❔

Removed the gasLimit override for payable functions.

## Why ❔

The explorer sets the gas limit to 10000000 by default for payable functions, preventing connected wallets like MetaMask from setting their own correct gas estimates.

### DIFF
--- a/packages/app/src/composables/useContractInteraction.ts
+++ b/packages/app/src/composables/useContractInteraction.ts
@@ -55,7 +55,6 @@ export default (context = useContext()) => {
         });
       const methodOptions = {
         value: ethers.utils.parseEther((params.value as string) ?? "0")
-        //gasLimit: "10000000",
       };
       const res = await method(
         ...[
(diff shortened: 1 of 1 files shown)


packages/app/src/composables/useContractInteraction.ts
```diff
@@ -54,8 +54,8 @@ export default (context = useContext()) => {
           return inputValue;
         });
       const methodOptions = {
-        value: ethers.utils.parseEther((params.value as string) ?? "0"),
-        gasLimit: "10000000",
+        value: ethers.utils.parseEther((params.value as string) ?? "0")
+        //gasLimit: "10000000",
```
Romsters, line 58: Let's just remove it please, don't see an obvious reason why it was there.

vasyl-ivanchuk:
Thanks for the PR, the same changes were already done in [this](https://github.com/matter-labs/block-explorer/pull/292) PR.

PR #299, opened 2024-10-23
feat: migrate to zksync-ethers

vasyl-ivanchuk:
# What ❔

Migrate the Explorer from the deprecated `zksync-web3` to `zksync-ethers`.

## Why ❔

Some explorer changes require SDK to be modified, in such cases the use of a deprecated version becomes a problem.

## Checklist


packages/api/src/common/transformers/hexToDecimalNumber.transformer.spec.ts
```diff
@@ -9,7 +9,7 @@ describe("hexToDecimalNumberTransformer", () => {
 
     it("returns hex representation of the decimal number string", () => {
       const result = hexToDecimalNumberTransformer.to("800");
```
JackHamer09, line 11: That's a weird transformer. Should ethers provide something like that already?
And also it's weird that the value changed. Was it incorrect before?

packages/api/test/transaction.e2e-spec.ts
```diff
@@ -130,11 +129,11 @@ describe("TransactionController (e2e)", () => {
...
-        gasPrice: BigNumber.from(1000 + i).toString(),
-        gasLimit: BigNumber.from(2000 + i).toString(),
-        maxFeePerGas: BigNumber.from(3000 + i).toString(),
-        maxPriorityFeePerGas: BigNumber.from(4000 + i).toString(),
-        gasPerPubdata: BigNumber.from(5000 + i).toHexString(),
+        gasPrice: BigInt(1000 + i).toString(),
+        gasLimit: BigInt(2000 + i).toString(),
+        maxFeePerGas: BigInt(3000 + i).toString(),
+        maxPriorityFeePerGas: BigInt(4000 + i).toString(),
+        gasPerPubdata: `0x${BigInt(5000 + i).toString(16)}`,
```
JackHamer09, line 136: Very minor thing, but maybe that should use either transformer if you decide to keep it, or an utility from ethers

packages/app/src/components/FeeData.vue
```diff
@@ -86,7 +84,7 @@ getTokenInfo(currentNetwork.value.baseTokenAddress);
 
 const initialFee = computed(() => {
   if (props.feeData) {
-    return BigNumber.from(props.feeData.amountPaid).add(props.feeData.amountRefunded).toHexString();
+    return `0x${(BigInt(props.feeData.amountPaid) + BigInt(props.feeData.amountRefunded)).toString(16)}`;
```
JackHamer09, line 87: Same as previous, I think you should just use toHex utility from ethers

packages/app/src/composables/useTransaction.ts
```diff
@@ -130,9 +129,9 @@ export default (context = useContext()) => {
           amountPaid: transactionDetails.fee.toString(),
           isPaidByPaymaster: false,
           refunds: [],
-          amountRefunded: BigNumber.from(0).toHexString(),
+          amountRefunded: `0x${BigInt(0).toString(16)}`,
```
JackHamer09, line 132: Same

packages/app/src/composables/useTransaction.ts
```diff
@@ -289,8 +288,8 @@ function mapTransfers(transfers: Api.Response.Transfer[]): TokenTransfer[] {
 }
 
 function sumAmounts(balanceChanges: TokenTransfer[]) {
-  const total = balanceChanges.reduce((acc, cur) => acc.add(cur.amount || 0), BigNumber.from(0));
-  return total.toHexString() as Hash;
+  const total = balanceChanges.reduce((acc, cur) => acc + BigInt(cur.amount || 0), BigInt(0));
+  return `0x${total.toString(16)}` as Hash;
```
JackHamer09, line 292: Same

packages/app/src/composables/useWallet.ts
```diff
@@ -0,0 +1,269 @@
...
+  isMetamaskInstalled: false,
+  isConnectPending: false,
+  isConnectFailed: false,
+
+  address: null,
+});
+
+export const isAuthenticated: RemovableRef<boolean> = useStorage<boolean>("useWallet_isAuthenticated", false);
+
+export default (
```
JackHamer09, line 66: I think it's a good practice to not export it as default but rather named export, so that it automatically shows up in vscode when you start typing the name

packages/app/src/utils/formatters.ts
```diff
@@ -67,9 +64,9 @@ export function formatHexDecimals(value: string, showValueAs: HexDecimals) {
   const validValue = value === "0x" ? "0" : value;
   const prefix = isHexString(validValue) ? "" : "0x";
   if (showValueAs === "Dec") {
-    return BigNumber.from(prefix + validValue).toString();
+    return BigInt(prefix + validValue).toString();
   }
-  return BigNumber.from(prefix + validValue).toHexString();
+  return `0x${BigInt(prefix + validValue).toString(16)}`;
```
JackHamer09, line 69: Same

packages/app/tests/utils/formatters.spec.ts
```diff
@@ -44,7 +44,7 @@ describe("formatters:", () => {
     );
   });
   it("returns formatted Hex data", () => {
-    expect(formatHexDecimals("32770", "Hex")).toBe("0x032770");
+    expect(formatHexDecimals("32770", "Hex")).toBe("0x32770");
```
JackHamer09, line 47: Was the formatting incorrect before?

packages/app/vite.config.ts
```diff
@@ -28,13 +29,13 @@ export default defineConfig({
           if (id.includes("@firebase")) {
             return "f";
           }
-          if (id.includes("@ethersproject")) {
+          if (id.includes("ethers")) {
```
JackHamer09, line 32: I think that now since you're importing only functions you use instead of whole ethers like before, maybe you don't need to split it in a separate chunk. But not 100% sure

JackHamer09 (changes requested):

packages/api/src/common/transformers/hexToDecimalNumber.transformer.spec.ts
vasyl-ivanchuk: Yeah, we need this one to transform decimal string to a hex string. Value has changed from `0x0320` to `0x320` which is the same value in the hex representation:
```
> BigInt("0x0320") === BigInt("0x320")
true
```
it's the same as with numbers in the decimal number system `800 === 0800`, so leading zeros can be ignored. Why the hex representation changed: old SDK uses `ethers v5` with `BigNumber` lib under the hood. In `ethers v6`, which is used by the new SDK, big numbers are now represented by the build in `bigint` type. `BigNumber` `toHexString` returned the value with one leading zero, while `bigint` `toString(16)` does not add any leading zeros.

packages/api/test/transaction.e2e-spec.ts
vasyl-ivanchuk: Agreed. Added an util function and used it in places like this.

packages/app/src/components/FeeData.vue
vasyl-ivanchuk: Ethers v6 doesn't provider a functiont transform a bigint to a hex string. Added an util function and used it in places like this.

packages/app/src/composables/useTransaction.ts
vasyl-ivanchuk: Added an util function and used it in places like this.

packages/app/src/composables/useTransaction.ts
vasyl-ivanchuk: Added an util function and used it in places like this.

packages/app/src/utils/formatters.ts
vasyl-ivanchuk: Added an util function and used it in places like this.

packages/app/tests/utils/formatters.spec.ts
vasyl-ivanchuk: It's the same value as leading zeros can be ignored. Explained [here](https://github.com/matter-labs/block-explorer/pull/299#discussion_r1819368551).

packages/app/src/composables/useWallet.ts
vasyl-ivanchuk: All existing composables in the project are exported as a default. I would rather keep this one in the same style, to keep things consistent.

JackHamer09 (approved):

PR #301, opened 2024-10-29
fix: configure ethers batching

vasyl-ivanchuk:
# What ❔

Add a configuration for ethers batching. Set the default batching config in case it's not defined.

## Why ❔

In ethers v6 batching is enabled by default and default values don't work well in our case. That's why we need a way to configure ethers batching and decrease the batching throughput by default.

## Checklist


packages/worker/.env.example
```diff
@@ -23,6 +23,9 @@ RPC_CALLS_DEFAULT_RETRY_TIMEOUT=30000
 RPC_CALLS_QUICK_RETRY_TIMEOUT=500
 RPC_CALLS_CONNECTION_TIMEOUT=20000
 RPC_CALLS_CONNECTION_QUICK_TIMEOUT=10000
+RPC_BATCH_MAX_COUNT=1
```
dutterbutter, line 26: Most people will make this the default, is `RPC_BATCH_MAX_COUNT=1` a good default to have?

packages/data-fetcher/.env.example
```diff
@@ -12,4 +12,8 @@ RPC_CALLS_RETRIES_MAX_TOTAL_TIMEOUT=120000
 RPC_CALLS_CONNECTION_TIMEOUT=60000
 RPC_CALLS_CONNECTION_QUICK_TIMEOUT=10000
 
+RPC_BATCH_MAX_COUNT=1
```
dutterbutter, line 15: Same question here? Are these good defaults? As most people will just make this their defaults.

packages/data-fetcher/src/config.spec.ts
```diff
@@ -21,6 +21,9 @@ describe("config", () => {
         rpcCallRetriesMaxTotalTimeout: 120000,
         rpcCallConnectionTimeout: 60000,
         rpcCallConnectionQuickTimeout: 10000,
+        rpcBatchMaxCount: 1,
```
dutterbutter, line 24: Do we have a spec test where this is not 1?

packages/worker/.env.example
vasyl-ivanchuk: Agreed, set batching size to 10 by default which should be a good default value.

packages/data-fetcher/src/config.spec.ts
vasyl-ivanchuk: Added tests to cover cases when newly added env vars set to a different values.

PR #308, opened 2024-11-05
fix: make tx receipt root field nullable

vasyl-ivanchuk:
# What ❔

Make tx receipt root field nullable.

## Why ❔

RPC doesn't return this field anymore. Here is the corresponding [PR](https://github.com/matter-labs/zksync-era/pull/3187).

## Checklist

### DIFF
--- a/packages/worker/src/entities/transactionReceipt.entity.ts
+++ b/packages/worker/src/entities/transactionReceipt.entity.ts
@@ -34,8 +34,8 @@ export class TransactionReceipt extends CountableEntity {
   @Column({ type: "int" })
   public readonly type: number;
 
-  @Column({ type: "bytea", transformer: hexTransformer })
-  public readonly root: string;
+  @Column({ type: "bytea", transformer: hexTransformer, nullable: true })
+  public readonly root?: string;
 
   @Column({ type: "varchar", length: 128 })
   public readonly gasUsed: string;
--- a/packages/worker/src/entities/transactionReceipt.entity.ts
+++ b/packages/worker/src/entities/transactionReceipt.entity.ts
@@ -34,6 +34,8 @@ export class TransactionReceipt extends CountableEntity {
   @Column({ type: "int" })
   public readonly type: number;
 
+  // this field is only relevant on L1 pre byzantium fork (EIP658),
+  // but was present in ZKsync API before, so it cannot be just dropped.
   @Column({ type: "bytea", transformer: hexTransformer, nullable: true })
   public readonly root?: string;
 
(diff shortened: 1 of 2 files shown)


packages/worker/src/entities/transactionReceipt.entity.ts
```diff
@@ -34,8 +34,8 @@ export class TransactionReceipt extends CountableEntity {
   @Column({ type: "int" })
   public readonly type: number;
 
-  @Column({ type: "bytea", transformer: hexTransformer })
-  public readonly root: string;
+  @Column({ type: "bytea", transformer: hexTransformer, nullable: true })
+  public readonly root?: string;
```
popzxc, line 38: Maybe makes sense leaving a comment that this field is only relevant on L1 pre byzantium fork ([EIP658](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-658.md)), but was present in ZKsync API before, so it cannot be just dropped.

packages/worker/src/migrations/1730806000905-TxReceiptRootColumnNullable.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class TxReceiptRootColumnNullable1730806000905 implements MigrationInterface {
+  name = "TxReceiptRootColumnNullable1730806000905";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "transactionReceipts" ALTER COLUMN "root" DROP NOT NULL`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "transactionReceipts" ALTER COLUMN "root" SET NOT NULL`);
```
popzxc, line 11: Will this work if there will already be null fields in the database? Shouldn't we have something like `DEFAULT '\\0x00..00'::bytea` or smth like this?

popzxc (dismissed):
LGTM as a quick fix, nits are up to you

packages/worker/src/migrations/1730806000905-TxReceiptRootColumnNullable.ts
vasyl-ivanchuk: Yes, it will, I checked it.

packages/worker/src/entities/transactionReceipt.entity.ts
vasyl-ivanchuk: good idea, done.

popzxc (approved):

PR #311, opened 2024-11-07
fix: apply proper type to each @ApiQuery decorator in the API contr…

RobbyUitbeijerse:
Hi!

The hosted OpenAPI spec that is abstracted from the `@ApiProperty` decorators using `@nestjs/swagger` is missing certain types. This is easy to check by simply opening the [OpenAPI spec](https://block-explorer-api.mainnet.zksync.io/docs-json) and checking the very first method and it's input parameters.

Screenshot 2024-11-07 at 16 31 07

The schema type the input parameter `address` is set to `{}` - this causes request client generators like [OpenAPI generator](https://github.com/OpenAPITools/openapi-generator) to interpret the parameter as an object, rather than the desired string.

To mitigate this, I simply added the `type` property in every `@ApiQuery` decorator instance, which should lead to proper output in the OpenAPI spec.

Let me know if you would like to see any changes here.

### DIFF
--- a/packages/api/src/api/api.controller.ts
+++ b/packages/api/src/api/api.controller.ts
@@ -84,6 +84,7 @@ export class ApiController {
   @ApiOperation({ summary: "Fetch the ABI for a given contract address" })
   @ApiQuery({
     name: "address",
+    type: String,
     description: "The contract address that has a verified source code",
     example: constants.verifiedContractAddress,
     required: true,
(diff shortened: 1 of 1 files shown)


vasyl-ivanchuk (approved):

vasyl-ivanchuk:
I changed the PR title to `fix` so that the new release version is created by the GitHub workflow. Otherwise, LGTM.

PR #314, opened 2024-11-11
fix: add missing swagger type annotations for BFF endpoints

vasyl-ivanchuk:
# What ❔

Add missing swagger type annotations for BFF endpoints.

## Why ❔

Auto code generators that work based on the Swagger docs rely on the specified Swagger type, and if it's missing, they assume the parameter is of the object type.

## Checklist

### DIFF
--- a/.github/workflows/app-e2e.yml
+++ b/.github/workflows/app-e2e.yml
@@ -2,10 +2,10 @@ name: BE App E2E tests
 
 on:
   workflow_call:
-    secrets:
-      ALLURE_TOKEN:
-        description: 'A token passed from the caller workflow'
-        required: true
+    # secrets:
+    #   ALLURE_TOKEN:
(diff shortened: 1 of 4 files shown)


.github/workflows/app-e2e.yml
```diff
@@ -2,10 +2,10 @@ name: BE App E2E tests
...
 on:
   workflow_call:
-    secrets:
-      ALLURE_TOKEN:
-        description: 'A token passed from the caller workflow'
-        required: true
+    # secrets:
+    #   ALLURE_TOKEN:
+    #     description: 'A token passed from the caller workflow'
+    #     required: true
```
vasyl-ivanchuk, line 8: Allure reports were disabled for the time being, so this variable is not used. However, since it's required, the workflow fails for forked PRs.

JackHamer09 (approved):

PR #324, opened 2024-11-23
feat: add nested routing

MexicanAce:
# What ❔

Add nested routing for tabs within tabs

## Why ❔

So we can provide links directly to sub-tabs. For example, I need a link directly to the `Read` tab of a verified contract, that would be: `https://explorer.zksync.io/address/0x000000000000000000000000000000000000800A#contract#read`

## Checklist

## Evidence

https://github.com/user-attachments/assets/6a3b4e6e-2e9e-47db-96f2-5a7e2e1ec161

### DIFF
--- a/packages/app/src/components/common/Tabs.vue
+++ b/packages/app/src/components/common/Tabs.vue
@@ -56,23 +56,41 @@ const props = defineProps({
     type: Boolean,
     default: true,
   },
+  hasNestedRoute: {
+    type: Boolean,
+    default: false,
+  },
 });
 
 const route = useRoute();
 const router = useRouter();
 
-const currentTabHash = ref(route?.hash && props.hasRoute ? route?.hash : props.tabs[0].hash);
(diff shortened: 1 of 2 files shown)


JackHamer09 (commented):
Maybe it could accept "routeHashNestingLevel" (0 by default and can be overriden to higher values)

MexicanAce:
> Maybe it could accept "routeHashNestingLevel" (0 by default and can be overriden to higher values)

Ideally yes, but because of how routing happens, we're only actually able to have 1 layer of nested `Tabs` components because if you move between the parent level `Tab`, the routing is lost/reset (You'll see that in the video when I move between parent level `Tabs` and go back, even though the previous sub-tab loads, the route is not updated)

vasyl-ivanchuk (approved):

vasyl-ivanchuk:
LGTM!

PR #337, opened 2024-11-29
feat: add support for evm-like contracts and transactions

tx-nikola:
# What ❔

Add badge that shows "EVM" for evm-like contracts and transactions.

## Checklist


Romsters (changes requested):
Please use `contractAddress` field to store deployed contract address similar to etherscan, set `to` to zero address.

petarTxFusion:
> Please use `contractAddress` field to store deployed contract address similar to etherscan, set `to` to zero address.

Its implemented

packages/api/src/common/transformers/normalizeAddress.transformer.ts
```diff
@@ -10,6 +10,9 @@ export const normalizeAddressTransformer: ValueTransformer = {
     if (!hex) {
       return null;
     }
+    if (hexTransformer.from(hex) === "0x") {
```
Romsters, line 13: Why? Empty string doesn't look like a valid address

packages/api/src/transaction/dtos/transaction.dto.ts
```diff
@@ -14,7 +14,7 @@ export class TransactionDto {
     description: "The address this transaction is to",
     example: "0xc7e0220d02d549c4846A6EC31D89C3B670Ebe35C",
   })
-  public readonly to: string;
+  public readonly to?: string;
```
Romsters, line 17: Why is it optional?

packages/api/src/transaction/entities/addressTransaction.entity.ts
```diff
@@ -19,8 +19,8 @@ export class AddressTransaction extends BaseEntity {
   @Column({ type: "bytea", transformer: hexTransformer })
   public readonly transactionHash: string;
 
-  @Column({ type: "bytea", transformer: normalizeAddressTransformer })
-  public readonly address: string;
+  @Column({ type: "bytea", transformer: normalizeAddressTransformer, nullable: true })
```
Romsters, line 22: why is it nullable?

packages/api/src/transaction/entities/addressTransaction.entity.ts
```diff
@@ -31,4 +31,7 @@ export class AddressTransaction extends BaseEntity {
 
   @Column({ type: "int" })
   public readonly transactionIndex: number;
+
+  @Column({ type: "boolean", nullable: true })
+  public readonly isEvmLike?: boolean;
```
Romsters, line 36: Does it have to be in this table?

packages/api/src/transaction/entities/transaction.entity.ts
```diff
@@ -31,8 +31,8 @@ export class Transaction extends BaseEntity {
   @Column({ generated: true, type: "bigint" })
   public number: number;
 
-  @Column({ type: "bytea", transformer: normalizeAddressTransformer })
-  public readonly to: string;
+  @Column({ type: "bytea", transformer: normalizeAddressTransformer, nullable: true })
+  public readonly to?: string;
```
Romsters, line 35: why optional?

packages/api/src/common/transformers/normalizeAddress.transformer.ts
0xVolosnikov: 0x00 and 0x are different values. Tx with empty `to` field is EVM deploy tx

packages/data-fetcher/src/transaction/transaction.service.ts
```diff
@@ -79,11 +82,24 @@ export class TransactionService {
       transactionReceipt
     );
 
+    const isEvmLike = transactionInfo.to === null;
+    const toAddress = isEvmLike ? "0x" : transactionInfo.to;
+    const contractAddress = isEvmLike
+      ? getCreateAddress({ from: transactionInfo.from, nonce: transactionInfo.nonce })
+      : "";
```
Romsters, line 89: Why empty string and not null?

packages/api/src/common/transformers/normalizeAddress.transformer.ts
0xVolosnikov: Then we should process 0x differently, earlier

Romsters: If you keep `getAddress(hexTransformer.from(hex))` wouldn't it just return a zero address?

petarTxFusion: No, it returns an error. I have left it like this in order to be aligned with etherscan since they are returning empty string in cases like this. We can switch to 0x000....

packages/api/src/transaction/dtos/transaction.dto.ts
petarTxFusion: Initially we where supporting to as a null when we started with evm support. I will change that back

packages/data-fetcher/src/transaction/transaction.service.ts
petarTxFusion: To align with etherscan where they are leaving empty string in cases like that one.

packages/worker/src/entities/addressTransaction.entity.ts
```diff
@@ -35,4 +35,7 @@ export class AddressTransaction extends BaseEntity {
 
   @Column({ type: "int" })
   public readonly transactionIndex: number;
+
+  @Column({ type: "boolean", nullable: true })
+  public readonly isEvmLike?: boolean;
```
vasyl-ivanchuk, line 40: I don't think this field should be here, as this entity is used to query transactions based on an address and join the corresponding transaction entities, which have all the fields available: https://github.com/matter-labs/block-explorer/blob/0b1a667f13fb4397c2fe78e401a77d4a0842e57c/packages/api/src/transaction/transaction.service.ts#L62

packages/api/src/common/transformers/normalizeAddress.transformer.ts
popzxc: I don't have much context on the PR, but from what I've understood `0x` is like of special case for contract deployment (like, and "empty to field"), which is different from 0x00..00 (IIRC in early times it was used for contract deployment, but nowadays you just don't specify anything).

packages/worker/src/entities/transaction.entity.ts
```diff
@@ -19,8 +19,8 @@ export class Transaction extends CountableEntity {
   @Column({ generated: true, type: "bigint" })
   public override readonly number: number;
 
-  @Column({ type: "bytea", transformer: hexTransformer })
-  public readonly to: string;
+  @Column({ type: "bytea", transformer: hexTransformer, nullable: true })
+  public readonly to?: string;
```
Romsters, line 23: Where is a migration to make this nullable?

packages/worker/src/migrations/1734453474315-MakeTransactionToFieldOptional.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class MakeTransactionToFieldOptional1734453474315 implements MigrationInterface {
+  name = "MakeTransactionToFieldOptional1734453474315";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "transactions" ALTER COLUMN "to" DROP NOT NULL`);
+    await queryRunner.query(`DROP INDEX "public"."IDX_f8b841e005e13e12008d9778ed"`);
```
Romsters, line 8: I think we can leave the index as it is and not recreate it.

Romsters:
Applied changes from https://github.com/matter-labs/block-explorer/pull/337/commits/e37496e0cec958c495fc18a60120ba0bb31c0c96.

Romsters (approved):

petarTxFusion (approved):

PR #347, opened 2024-12-04
fix: pass arguments for the contract functions in the correct order

tx-nikola:
# What ❔
This PR fixes the bug where the contract function arguments are not passed in the correct order.

## Why ❔

In order to call the contract functions correctly, we need to pass the arguments in the right order.

# Explanation
The issue was that we were trying to get `Object.entries(params)` without actually sorting it, so, for example, contract `[0xE1D6A50E7101c8f8db77352897Ee3f1AC53f782B](https://explorer.zksync.io/address/0xE1D6A50E7101c8f8db77352897Ee3f1AC53f782B#contract#write-proxy)`, the `revokeRole` function asks for `[role, account]` arguments to be passed.
In the form object, it looks like this:
```
{
  account: "0xExample",
  role: "0xExample
}
```
and when we run `Object.entries(params)` we always get it like this:
```
[
  ["account", "0x969Bb8Ae65602B4F8f9B459a11084e591c4491C7"],
  ["role", "0x03c321230b026fb61a5a21f62c5f618751ec6d8435327f673bae5bfa570e5879"]]
]
```
and later after we map it, it becomes `["0x969Bb8Ae65602B4F8f9B459a11084e591c4491C7", "0x03c321230b026fb61a5a21f62c5f618751ec6d8435327f673bae5bfa570e5879"]` - swapping the arguments from their intended places.

This issue was happening when someone types the account first, therefore forms become
```
{
  account: "0xExample",
  role: "0xExample
}
```
instead of 
```
{
  role: "0xExample,
  account: "0xExample"
}
```
thus messing up the order of the arguments in the function call.

## Checklist


vasyl-ivanchuk (approved):

vasyl-ivanchuk:
@tx-nikola looks good to me! I did a minor [improvement](https://github.com/matter-labs/block-explorer/pull/347/commits/27ed9d7e846985160826fd7ff8de0e3360cc3add) and approved the PR.

Issue #287, opened 2024-09-24
Add token bridge details to tokens


popzxc:
#### 📝 Description

Right now, the same token can exist both on the default bridge and on a specific bridge.
Example would be WETH, right now on the [token page](https://explorer.zksync.io/tokens) you can see two WETH tokens:
- 0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91 
- 0xf00DAD97284D0c6F06dc4Db3c32454D4292c6813 

Right now it's not clear which one should be used.
What we can do is to add labels, e.g. "Default bridge" or "L2-native", which will provide more context on hover.

Similar case exists for wstETH.

#### 🤔 Rationale

Having two versions of token without additional context is confusing.

PatrickAlphaC:
Bump. This took me a long time. Also, different projects (like aave) use different tokens. For WETH, it's odd that the WETH with the deposit function is not used.

PR #360, opened 2024-12-16
feat: add lables for bridged and native tokens

kiriyaga-txfusion:
# What ❔

Add labels for bridged and native tokens.

## Why ❔

The same token can exist both on the default bridge and on a specific bridge. To help users differentiate between them, we have added labels.

image

fixes https://github.com/matter-labs/block-explorer/issues/287

## Checklist

- [+] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [+] Tests for the changes have been added / updated.

### DIFF
--- a/packages/app/src/components/TokenIconLabel.vue
+++ b/packages/app/src/components/TokenIconLabel.vue
@@ -17,12 +17,22 @@
         />
       </div>
     </AddressLink>
-    <div class="token-info" v-if="name && symbol">
-      <div class="token-symbol">
-        {{ symbol }}
-      </div>
-      <div class="token-name">
-        {{ name }}
+    <div class="token-info-container">
+      <div class="token-info" v-if="name && symbol">
+        <div class="token-symbol">
+          {{ symbol }}
+        </div>
+        <div class="token-name">
+          {{ name }}
+        </div>
+        <div class="token-badge">
+          <Badge v-if="bridged" color="primary" class="verified-badge" :tooltip="t('tokensView.table.bridged.tooltip')">
+            {{ t("tokensView.table.bridged.title") }}
+          </Badge>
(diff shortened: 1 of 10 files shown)


packages/app/src/components/TokenIconLabel.vue
```diff
@@ -17,12 +17,22 @@
...
-        {{ name }}
+    <div class="token-info-container">
+      <div class="token-info" v-if="name && symbol">
+        <div class="token-symbol">
+          {{ symbol }}
+        </div>
+        <div class="token-name">
+          {{ name }}
+        </div>
+        <div style="display: flex; gap: 0.5rem">
```
Romsters, line 28: please move inline styles to css class

packages/app/src/locales/en.json
```diff
@@ -591,7 +591,15 @@
         "table": {
             "tokenName": "Token Name",
             "price": "Price",
-            "tokenAddress": "Token Address"
+            "tokenAddress": "Token Address",
+            "bridged": {
+                "title": "Default bridge",
```
Romsters, line 596: What if for some tokens it's not the default bridge? Maybe better to say just `Bridged`?

packages/app/src/components/TokenIconLabel.vue
Romsters, line 28: Would also suggest to display it in a separate column called `Type` or `Bridge`, would look a lot better as for me.

vasyl-ivanchuk: Given that now we also have the _L1 token address_ column, I reverted back to badges because, with a separate column for the type, the table looks overwhelmed. Here is how it looks:
image

vasyl-ivanchuk (approved):

Issue #351, opened 2024-12-04
Address explorer etherscan-like APIs incompatibilities


Romsters:
#### 📝 Description

Address explorer's [verification etherscan-like API endpoint](https://block-explorer-api.mainnet.zksync.io/docs) incompatibilities to make verification work with foundry.

#### 🤔 Rationale

In order for contract verification through foundry the API endpoint for verification should work the same way.

#### 📋 Additional Context

Figure out what are the discrepancies and address them.

PR #367, opened 2024-12-25
fix: verification endpoint alignment with etherscan

kiriyaga-txfusion:
# What ❔

Resolved incompatibilities with the explorer's verification API endpoint.

## Why ❔

For contract verification through Foundry and Hardhat, the API endpoint should function in the same way as it works on Etherscan.

This PR resolves the issue:
https://github.com/matter-labs/block-explorer/issues/351

## Checklist

- [+] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [+] Tests for the changes have been added / updated.

### DIFF
--- a/packages/api/src/api/contract/contract.controller.spec.ts
+++ b/packages/api/src/api/contract/contract.controller.spec.ts
@@ -278,7 +278,7 @@ describe("ContractController", () => {
         contractname: "contracts/HelloWorld.sol:HelloWorld",
         compilerversion: "0.8.17",
         optimizationUsed: "1",
-        zkCompilerVersion: "v1.3.14",
+        zksolcVersion: "v1.3.14",
         constructorArguements: "0x94869207468657265210000000000000000000000000000000000000000000000",
         runs: 700,
         libraryname1: "contracts/MiniMath.sol:MiniMath",
@@ -426,7 +426,7 @@ describe("ContractController", () => {
         contractname: "contracts/Greeter.vy:Greeter",
         compilerversion: "0.3.3",
         optimizationUsed: "1",
-        zkCompilerVersion: "v1.3.11",
+        zksolcVersion: "v1.3.11",
       } as unknown as VerifyContractRequestDto;
 
       pipeMock.mockReturnValue(
@@ -442,7 +442,7 @@ describe("ContractController", () => {
         codeFormat: "vyper-multi-file",
         compilerVyperVersion: "0.3.3",
         compilerZkvyperVersion: "v1.3.11",
-        constructorArguments: undefined,
+        constructorArguments: null,
         contractAddress: "0x14174c76E073f8efEf5C1FE0dd0f8c2Ca9F21e62",
         contractName: "contracts/Greeter.vy:Greeter",
         optimizationUsed: true,
--- a/packages/api/src/api/contract/contract.controller.spec.ts
+++ b/packages/api/src/api/contract/contract.controller.spec.ts
@@ -442,7 +442,7 @@ describe("ContractController", () => {
         codeFormat: "vyper-multi-file",
         compilerVyperVersion: "0.3.3",
         compilerZkvyperVersion: "v1.3.11",
-        constructorArguments: null,
+        constructorArguments: "0x",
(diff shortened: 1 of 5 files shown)


packages/api/src/api/contract/contract.controller.ts
```diff
@@ -136,6 +135,12 @@ export class ContractController {
       ContractVerificationCodeFormatEnum.solidityJsonInput,
     ].includes(request.codeformat);
 
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const semver = require("semver");
```
Romsters, line 139: Why not import it?

packages/api/src/api/dtos/contract/verifyContractRequest.dto.ts
```diff
@@ -115,19 +117,19 @@ export class VerifyContractRequestDto {
...
-  @IsNotEmpty({ message: "Missing optimizationUsed" })
+  @IsOptional()
   public optimizationUsed: string;
 
   @ApiProperty({
     name: "constructorArguements",
     description: "Contract constructor arguments",
     example:
-      "0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000094869207468657265210000000000000000000000000000000000000000000000",
+      "0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000094869207468657265210000000000000000000000000000000000000000000000 or 000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000094869207468657265210000000000000000000000000000000000000000000000",
```
Romsters, line 132: What's that `or` for? If this is to show multiple examples just use `examples` field, it's array.

packages/api/src/api/contract/contract.controller.ts
kiriyaga-txfusion: Import doesn't work because the semver package only functions with CommonJS modules. I found a workaround using require to properly use the object.

packages/api/src/api/dtos/contract/verifyContractRequest.dto.ts
kiriyaga-txfusion: It shows that constructor arguments can be sent either with or without the 0x prefix, as the Etherscan API only accepts without the prefix.

packages/api/src/api/contract/contract.controller.spec.ts
```diff
@@ -278,7 +278,7 @@ describe("ContractController", () => {
         contractname: "contracts/HelloWorld.sol:HelloWorld",
         compilerversion: "0.8.17",
         optimizationUsed: "1",
-        zkCompilerVersion: "v1.3.14",
+        zksolcVersion: "v1.3.14",
```
popzxc, line 281: Does this mean that the change is breaking? If so, can we make it backward-compatible (e.g. add support for property alias)?

packages/api/src/api/dtos/contract/verifyContractRequest.dto.ts
Romsters: Then use array with both these values please with and without 0x.

packages/api/src/api/contract/contract.controller.ts
Romsters: Did you try to add [types](https://www.npmjs.com/package/@types/semver)?

packages/api/src/api/contract/contract.controller.spec.ts
Romsters: Yes, this change is breaking. TBH I would better leave `zkCompilerVersion` for consistency and ask etherscan to rename on their side, because for solc it's not `solcVersion`, it's `compilerversion`. If we still decide to go with `zksolcVersion` then we should support both `zksolcVersion` and `zkCompilerVersion` to avoid breaking changes.

packages/api/src/api/contract/contract.controller.ts
kiriyaga-txfusion: I tried with `semver` and its type, but the dependency object is always null.

Issue #370, opened 2025-01-13
Missing blocks in the DB


vasyl-ivanchuk:
#### 📝 Description

At any point in time, for any saved block in the DB, prior blocks (and their corresponding information) should already be present in the DB. This is a design decision, and some functionality depends on it. It has been noticed that there are missing blocks in the DB, which is unexpected.

#### 🔄 Reproduction Steps

I haven't been able to reproduce the problem yet, but there are a few environments where I see the symptoms.

#### 🤔 Expected Behavior

There are no missing blocks in the DB. At any point in time, if there are saved blocks in the DB, the following SQL query must return true:
```
SELECT COUNT(number) = MAX (number) + 1	FROM public.blocks -- +1 for the block #0
```

#### 😯 Current Behavior

There are missing blocks in the DB. For example (blocks are ordered by number ASC):
```
...
number	timestamp	createdAt	updatedAt
12237	2024-10-29T08:09:52.000Z	2024-10-29T08:14:53.615Z	2024-10-29T08:14:53.615Z
12248	2024-10-29T08:59:58.000Z	2024-10-29T16:17:14.952Z	2024-10-29T16:17:14.952Z
12257	2024-10-29T09:43:19.000Z	2024-10-29T16:17:15.018Z	2024-10-29T16:17:15.018Z
...
```

#### 🖥️ Environment

It's been noticed on a few environments, DM me for more details.

vasyl-ivanchuk:
I've checked block processing and the revert functionality in the worker service and haven’t noticed anything that could have caused the issue. The way the DAL uses the DB (transactions, isolation levels, etc.) looks good as well. I’ve added an alert for missing blocks, so next time it happens, we will be notified and can react immediately, having more data to analyze (logs, etc.).

For now, I’m closing the issue. If needed, a new one will be created.

PR #371, opened 2025-01-14
Add metric for missing blocks

vasyl-ivanchuk:
# What ❔

Add missing blocks metric.

## Why ❔

To detect when there are missing blocks in the DB. Issue with missing blocks is described [here](https://github.com/matter-labs/block-explorer/issues/370).

## Checklist


packages/app/tests/e2e/features/redirection/redirectionSet1.feature
```diff
@@ -21,13 +21,13 @@ Feature: Redirection
...
+    Then New page address matches the "<regexp>"
 
     Examples:
-      | Icon    | url                                |
+      | Icon    | regexp                                |
       # discord renamed to "join"
-      | join    | https://join.zksync.dev/           |
-      | twitter | https://x.com/zksync               |
+      | join    | ^https://join.zksync.dev/$           |
+      | x.com | ^https://x.com/zksync(\\?.*)?$               |
```
vasyl-ivanchuk, line 30: X redirects to `https://x.com/zksync` with a query string, since it's not clear what that query string means and if it will change or not I decided to check based on the RegExp.

Romsters (approved):

Issue #352, opened 2024-12-04
Add token page


Romsters:
#### 📝 Description

Add token page similar to [etherscan](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7) 

#### 🤔 Rationale

Currently this page is missing, this will be very useful to have a page with all the token details and stats.

PR #373, opened 2025-01-17
feat: add token page with holders list

tx-nikola:
# What ❔

Add the token page, so that the ERC20 tokens are represented similarly to how Etherscan shows them.
The route for token page is `/token/{address}`, instead of `/account/{address}`.

## Checklist

This PR fixes https://github.com/matter-labs/block-explorer/issues/352


packages/api/src/balance/balance.service.ts
```diff
@@ -99,4 +103,71 @@ export class BalanceService {
     const balancesRecords = await balancesQuery.getMany();
     return balancesRecords;
   }
+
+  public async getBalancesForTokenAddress(
```
Romsters, line 107: This query will take a while on mainnet. To improve it an index ("tokenAddress", "address", "blockNumber") can be added, but even with that it'll take a lot of time for tokens with large number of holders.

kiriyaga-txfusion: I’m using the logic from another fetch that retrieves balances, but I’ve adjusted it for a different use case with the token address. Do you have any suggestions, like a caching mechanism or something similar?

Romsters: Caching won't help, DB schema has to be redesigned to make the queries simpler.

PR #378, opened 2025-01-22
fix: handle potential concurrent system contracts insertion

Romsters:
If indexer finds a contract deployment event for a system contract at the same time as system contracts service inserts it - there will be a collision. It can very likely happen especially after these events are added on the blockchain side for initial deployments of system contracts.

### DIFF
--- a/packages/worker/src/contract/systemContract.service.ts
+++ b/packages/worker/src/contract/systemContract.service.ts
@@ -26,7 +26,7 @@ export class SystemContractService {
         const bytecode = await this.blockchainService.getCode(contract.address);
         // some contract might not exist on the environment yet
         if (bytecode !== "0x") {
-          await this.addressRepository.add({
+          await this.addressRepository.upsert({
             address: contract.address,
             bytecode,
           });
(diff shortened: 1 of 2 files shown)


packages/worker/src/contract/systemContract.service.ts
```diff
@@ -26,7 +26,7 @@ export class SystemContractService {
         const bytecode = await this.blockchainService.getCode(contract.address);
         // some contract might not exist on the environment yet
         if (bytecode !== "0x") {
-          await this.addressRepository.add({
+          await this.addressRepository.upsert({
```
vasyl-ivanchuk, line 29: @Romsters in the case you've described it's unlikely that the bytecodes will be different, but still I'd rather wrap `add` with try / catch or use `add` with `on conflict do nothing` not to overwrite the bytecode with an obsolete value.

Romsters: It will not update the bytecode in that case as it has block number condition that checks if the data is not outdated before updating

vasyl-ivanchuk (approved):

Issue #346, opened 2024-12-04
feature request: Support verification of EVM contracts


0xVolosnikov:
#### 📝 Description

Requires https://github.com/matter-labs/block-explorer/issues/318 to be done.

We need to support verification of EVM contracts. 

#### 🤔 Rationale

Block explorer should support all types of contracts that are used in our system.

#### 📋 Additional Context

EVM verification is supported already in contract_verifier crate of zksync-era: https://github.com/matter-labs/zksync-era/pull/3225

PR #391, opened 2025-03-13
feat: add EVM verification

petarTxFusion:
## Checklist

This PR fixes: https://github.com/matter-labs/block-explorer/issues/346


packages/app/src/composables/useContractVerification.ts
```diff
@@ -86,32 +86,55 @@ export default (context = useContext()) => {
         ContractVerificationCodeFormatEnum.soliditySingleFile,
         ContractVerificationCodeFormatEnum.solidityMultiPart,
       ].includes(data.codeFormat);
-
-      const { sourceCode, zkCompilerVersion, compilerVersion, ...payload } = data;
+      const { sourceCode, zkCompilerVersion, compilerVersion, isEVM, ...payload } = data;
       const response = await $fetch(`${context.currentNetwork.value.verificationApiUrl}/contract_verification`, {
         method: "POST",
         body: {
```
popzxc, line 92: Would it make sense to split it into EVM and non-EVM parts? Right now it feels a bit too boilerplate-ish and it's hard to understand what's going on.

petarTxFusion: It makes sense, I will just separate it in two different calls

packages/app/src/components/common/CheckBoxInput.vue
```diff
@@ -35,8 +39,14 @@ const inputted = computed({
...
+    @apply opacity-50;
+  }
+
   input {
-    @apply rounded me-1 text-primary-600 cursor-pointer border-neutral-200 checked:border-primary-600 hover:border-primary-600 ring-transparent;
+    @apply rounded me-1 text-primary-600 cursor-pointer border-neutral-200 checked:border-primary-600 disabled:opacity-50 hover:border-primary-600 ring-transparent;
   }
 }
+.checkbox-input-container {
+}
```
vasyl-ivanchuk, line 51: empty CSS class

packages/app/src/components/common/CheckBoxInput.vue
```diff
@@ -35,8 +39,14 @@ const inputted = computed({
 .checkbox-input-container {
   @apply flex items-center mb-1 cursor-pointer float-right leading-snug;
 
+  &.disabled {
+    @apply opacity-50;
```
vasyl-ivanchuk, line 43: When the input is disabled, we also need to redefine the hover state. Right now, even if the control is disabled and you hover over the checkbox, the box is highlighted with borders:
image

image

packages/app/src/views/ContractVerificationView.vue
```diff
@@ -520,10 +568,18 @@ const v$ = useVuelidate(
...
+  selectedCompilerVersion.value = EVM_VERSION_TO_TARGET[0] || "";
+}
+
 function onZkVMSelectionChanged() {
   selectedCompilerVersion.value = selectedCompiler.value.versions[0] || "";
 }
 
+function onEVMSelectionChanged() {
+  selectedCompilerVersion.value = EVM_VERSION_TO_TARGET[0] || "";
+}
```
vasyl-ivanchuk, line 581: onEVMSelectionCustomChanged and onEVMSelectionChanged are identical, probably the wrong value is reset in one of them.

packages/app/src/views/ContractVerificationView.vue
vasyl-ivanchuk, line 1: You're adding new controls but binding them to the same model. Both the `zkCompilerVersion` and `evmVersionToTarget` dropdowns are bound to the `selectedZkCompilerVersion` model, which:

1. is hardher to understand.
2. creates a mess with the shared state between the controls. For instance, do the following:
- open the view and select `Zksolc version`
- select the `EVM` checkbox
- the `Zksolc version` dropdown is changed to `EVM Version to target` but the value still remains:
image

Instead of using the same model, I would suggest adding a new field for `evmVersion`. Then when constructing the request to send for verification, you can decide which field to read based on the selected EVM setting.

Apart from that, is it correct that when I select the `EVM` checkbox, a new item called `default` is added to the `Solc Version` dropdown? If so, why does it disappear when I select any non-default value?

Please check the different states of the form and ensure that the values are properly populated and reset when expected.

0xVolosnikov:
I don't see "runs" param here for Solc. It is important - Solc will produce different bytecodes with different `runs`

packages/app/src/components/common/CheckBoxInput.vue
tx-nikola: Fixed.

packages/app/src/views/ContractVerificationView.vue
tx-nikola: Removed the same functions.

packages/app/src/components/common/CheckBoxInput.vue
tx-nikola: Removed.

packages/app/src/components/common/CheckBoxInput.vue
vasyl-ivanchuk: Please, also add `cursor-default` for both `checkbox-input-container &.disabled` and `checkbox-input-container &.disabled input`. Otherwise the pointer cursor is displayed on hover even if the input is disabled.

packages/app/src/views/ContractVerificationView.vue
tx-nikola: I separated the state - now they're separate and both are being passed as the form values, together with the isEVM check; also fixed the bug where the "default" value would be added to the wrong dropdown as the default value.

packages/app/src/locales/en.json
```diff
@@ -494,7 +512,12 @@
                     "required": "Zkvyper version is required"
                 },
                 "error": "Unable to get list of supported Zkvyper versions",
-                "details": "Details"
+                "details": "Details",
+                "evmVersionToTarget": {
+                    "label": "EVM Version to target",
+                    "custom": "Custom",
+                    "inputPlaceholder": "Custom version..."
+                }
```
vasyl-ivanchuk, line 520: It seems that only a few of the new translations are being used. Let's remove the unused ones.

Please change the text from `EVM Version to target` to `EVM Version`, as it's clear what is meant here and the longer text overflows with other inputs on small resolutions.

0xVolosnikov:
Couple of thoughts from my side:

- 200 should be default value for runs if optimization enabled (default in foundry/hardhat)
- I think "Cancun" EVM version should be used by default. It is unclear what "default" option means
- I guess "ZkVM" mode should be checked by default, otherwise it is unclear what happens by default

0xVolosnikov:
Also as I see "Contract Path" disappears for some reason in EVM mode

packages/app/src/composables/useContractVerification.ts
```diff
@@ -86,32 +86,49 @@ export default (context = useContext()) => {
...
-            ...(typeof sourceCode !== "string" && {
-              sourceCode: Object.keys(sourceCode.sources).reduce((acc: Record<string, string>, filename: string) => {
-                acc[filename.replace(".vy", "")] = sourceCode.sources[filename].content;
-                return acc;
-              }, {}),
-            }),
-            compilerZkvyperVersion: zkCompilerVersion,
-            compilerVyperVersion: compilerVersion,
-          }),
+          sourceCode: sourceCodeVal,
```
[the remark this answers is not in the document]
tx-nikola: Fixed.

tx-nikola:
@0xVolosnikov  Should be fixed now.

0xVolosnikov:
I think ZkVM and EVM options are mutually exclusive and we should have just a simple switch between them. Rn you need to uncheck ZkVM (and after that you are in some strange state) and separately check EVM

0xVolosnikov:
And "default" EVM Version is still confusing, it is unclear what does it mean. I'd prefer "Cancun (default)" or smth

0xVolosnikov:
> I think ZkVM and EVM options are mutually exclusive and we should have just a simple switch between them. Rn you need to uncheck ZkVM (and after that you are in some strange state) and separately check EVM

Ah, "zkVM" is about our fork of Solc, not verification for zkVM. This is very confusing actually.

0xVolosnikov:
And imo to switch between EVM and EraVM modes it is better to use some kind of toogle switch instead of checkbox

tx-nikola:
1. Separate `zkVM` and `EVM` as they handle separate logic
2. Rename `zkVM` label to `Use zkVM solc`
3. When `EVM` checkbox is **not** checked, it means that `EraVM` mode is active? So it's either `EVM` or `EraVM` options, they can't both be active, and at least one of them is active always
4. Therefore, instead of `EVM` being a checkbox, we should modify it so that it's a toggle switch, one side being `EraVM`, and other side being `EVM`, something like this: "EraVM [checkbox component] EVM"
5. For the default version in the EVM version dropdown, set it as `cancun` instead of `default` - should we remove the "default" option completely? I used the default only because I saw it on the Etherscan Contract Verification as one of the options in their dropdown

@0xVolosnikov  Let me know if these changes are correct and if I'm not mistaken somewhere, so we can quickly fix it 🙌

0xVolosnikov:
Yep, you are right - current "zkVM" checkbox is about specialized compiler version and "EVM" checkbox is about EVM/zkVM compilation **target**. So "EVM" would a lot more natural as toggle switch (EVM vs zkVM, or EraVM, whatever).

About "default" option - I see. In Etherscan it is tied to Solc versions (so each version has default EVM version). So let's keep "default" option.

tx-nikola:
@0xVolosnikov I pushed the changes, let me know if it works for you, or if it would be better in a different place... It's kinda hard to think of a place where the toggle would "fit" :)

0xVolosnikov:
Some test are failing

tx-nikola:
Fixed it

0xVolosnikov:
1. I think "compilation target" should be before "compilation info". 
2. Now we don't need "Use zkVM solc" enabled by default, sorry, it was some misunderstanding from my side

Otherwise lgtm

tx-nikola:
Updated it - re-ordered the components so that compolation target is before the compilation info; and now "Use zkVM solc" is not checked by default

0xVolosnikov:
Nice, I like it!

@vasyl-ivanchuk @popzxc PTAL

tx-nikola:
@vasyl-ivanchuk Fixed - let me know if it's okay.

vasyl-ivanchuk:
@tx-nikola one thing left: when the optimization is disabled, `runs` is still added to the `optimizer` object. It's either default (200) or a value from the previous selection:
```
{
    "language": "Solidity",
    "sources": ...,
    "settings": {
        "optimizer": {
            "enabled": false,
            "runs": 500
        }
    }
}
```

tx-nikola:
Fixed - now `runs` is passed only if the target is EVM and the optimization is set to "Yes".

vasyl-ivanchuk (approved):

vasyl-ivanchuk:
LGTM!

Issue #386, opened 2025-02-27
Paging component is broken on the blocks page


Romsters:
#### 📝 Description

Paging component styles are broken on the blocks page.

https://explorer.zksync.io/blocks/

PR #392, opened 2025-03-13
feat: fix paging component alignment on blocks view page

tx-nikola:
# What ❔

The pagination component had both the dropdown and pagination stacked one under another.
Removing the parent wrapper fixed the alignment.

This PR fixes https://github.com/matter-labs/block-explorer/issues/386


vasyl-ivanchuk:
With the paging component fixed and properly displayed, it is now visible that the page size part doesn't work on this view. But this issue is not related to the changes in this PR, as page size is also broken on a few other views. I will create a separate issue for it.

LGTM!

vasyl-ivanchuk (approved):

PR #410, opened 2025-04-30
feat: add first iteration of login view

Jrigada:
# What ❔
- Migrate logic for login and context from DoubleZero
- Add first iterarion of login functionality and navigation guard

## Why ❔
- Add login functionality to block explorer

image

image

image

image

## Checklist


packages/app/src/App.vue
```diff
@@ -1,12 +1,19 @@
 <template>
   <template v-if="isReady">
-    <the-header :class="$route?.name" />
-    <div class="container-app">
+    <the-header
+      v-if="!['login', 'reviewing-permissions', 'not-authorized'].includes(String($route.name || ''))"
```
aon, line 4: Perhaps it's better if we check specific routes, otherwise we might match other unintended paths when using "includes" instead of something like "startsWith".

packages/app/src/components/ConnectMetamaskButton.vue
```diff
@@ -60,34 +167,45 @@ const buttonText = computed(() => {
   }
   return t("connectMetamaskButton.label");
 });
+
+const shortenedAddress = computed(() => {
+  if (!displayAddress.value) return "-";
+  return `${displayAddress.value.slice(0, 6)}...${displayAddress.value.slice(-4)}`;
+});
 </script>
```
aon, line 175: This could be moved into utils/formatters

packages/app/src/components/WalletStatusBar.vue
```diff
@@ -0,0 +1,88 @@
...
+const context = useContext();
+const { logout } = useLogin(context);
+const currentNetwork = computed(() => context.currentNetwork.value);
+
+const { address, getEthereumProvider, isReady } = useWallet({
+  ...context,
+  currentNetwork: computed(() => ({
+    explorerUrl: context.currentNetwork.value.rpcUrl,
+    chainName: context.currentNetwork.value.l2NetworkName,
+    l1ChainId: null as unknown as number,
```
calvogenerico, line 27: Is it ok to have the contstant null here?

Also, typescript it's a little bit special with the `as`. Even when this type, I feel that sending "null" as a number might cause issues down the road.

packages/app/src/components/modals/WalletInfoModal.vue
```diff
@@ -0,0 +1,160 @@
...
+
+defineEmits<{
+  (eventName: "close"): void;
+  (eventName: "disconnect"): void;
+}>();
+
+const { t } = useI18n();
+
+const switchNetwork = async () => {
+  // TODO: Implement network switch
```
calvogenerico, line 83: Do we have a plan for this? If we want a temporal solution maybe we can open an alert saying "comming soon" or something like that

packages/app/src/composables/useEnvironmentConfig.ts
```diff
@@ -42,5 +42,11 @@ export default () => {
         ? config.value.networks.filter((e) => e.published === true)
         : []
     ),
+    isPrividium: computed((): boolean => {
+      return config.value?.networks?.[0]?.isPrividium ?? false;
+    }),
+    baseTokenAddress: computed(
+      () => config.value?.networks?.[0]?.baseTokenAddress ?? "0x000000000000000000000000000000000000800A"
```
calvogenerico, line 49: Can we make this a constant somewhere? I mean the hex value

packages/app/src/composables/useWallet.ts
```diff
@@ -31,6 +31,7 @@ type UseWallet = ToRefs<WalletState> & {
 
   getL1Signer: () => Promise<L1Signer>;
   getL2Signer: () => Promise<Signer>;
+  getEthereumProvider: () => Promise<BaseProvider | undefined>;
```
calvogenerico, line 34: Even when nullables are undefinables are kind of the same, I believe that using `BaseProvider | null` it's a little bit better. The main reason is that null denotates intention, while undefined can totally be a mistake like a variable not initialized or accessing an empty property of an object.

packages/app/src/views/LoginView.vue
```diff
@@ -0,0 +1,65 @@
...
+  } catch (error) {
+    console.error("Login failed:", error);
+  }
+};
+
+watch(
+  () => context.user.value.loggedIn,
+  (isLoggedIn) => {
+    if (isLoggedIn) {
+      const redirectPath = route.query.redirect as string;
```
calvogenerico, line 52: Why do you need the `as string` here?
I feel that if this can be undefined or something that is not string there should be a validation or an explicit check here.

packages/app/src/components/WalletStatusBar.vue
Jrigada: Yes I agree, I didn't want to make it optional in the type so maybe passing a value that makes sense here and is not bypassing typescript type checking could be a better idea.

packages/app/src/components/modals/WalletInfoModal.vue
Jrigada: We have a plan, it should be integrated by today max tomorrow.

PR #426, opened 2025-05-15
feat: add prividium private rpc token

aon:
# What ❔

## Checklist


calvogenerico:
At some point before the service starts it should explode if `prividium` is active and the URL is not present.

calvogenerico:
Same for the RPC secret

packages/api/src/auth/auth.controller.ts
```diff
@@ -111,6 +111,31 @@ export class AuthController {
...
+    const response = await fetch(this.configService.get("prividiumPrivateRpcUrl"), {
+      method: "POST",
+      body: JSON.stringify({
+        address: req.session.siwe.address,
+        secret: this.configService.get("prividiumPrivateRpcSecret"),
+      }),
+      headers: { "Content-Type": "application/json" },
+    });
+    const data = await response.json();
+    return data;
```
calvogenerico, line 136: What do you think about enforcing the response format here?

```
return {
  token: data.token,
  ok: data.ok
}
```

Another option would be parse the response and throe 424 or 500 if response doesn't follow the expected format.

calvogenerico (approved):

PR #428, opened 2025-05-16
feat: dz phase 1 coverage

calvogenerico:
# What ❔

100% coverage for new features, including dz new code.


packages/api/src/middlewares/prividium-filtering.middleware.ts
```diff
@@ -40,7 +44,7 @@ export class PrividiumFilteringMiddleware implements NestMiddleware {
     const isContract = !!(addressRecord && addressRecord.bytecode.length > 2);
 
     if (!reqAddress) {
-      throw new ForbiddenException({ message: "Access denied" });
+      return;
     }
```
aon, line 48: Will this return a forbidden http error?

calvogenerico: The api `/address` doesnt actually exist. This is going to continue with the normal flow and eventually return a 404.

aon (approved):

PR #435, opened 2025-05-19
feat: generate siwe on server

aon:
# What ❔

## Checklist


packages/api/src/cookie-session.d.ts
```diff
@@ -6,6 +6,7 @@ declare global {
     interface CookieSessionObject {
       nonce?: string;
```
calvogenerico, line 7: I feel that nonce is kind of redundant now, right?

packages/api/src/auth/auth.controller.ts
```diff
@@ -44,6 +44,35 @@ export class AuthController {
...
+  @Header("Content-Type", "text/plain")
+  @ApiOkResponse({
+    description: "Message was returned successfully",
+    schema: { type: "string" },
+  })
+  @ApiBadRequestResponse({
+    description: "Nonce must be requested first",
+    schema: { type: "object", properties: { message: { type: "string" } } },
+  })
+  public getMessage(@Req() req: Request, @Body() body: { address: string }): string {
```
calvogenerico, line 57: Does this means that now the `app` has to call `GET /nonce` and then `POST /message`? Maybe we should unify those 2.

aon: Definitely. Working on it.

calvogenerico (approved):

calvogenerico (approved):

PR #453, opened 2025-05-29
fix: low performing transaction find all query in public mode

aon:
# What ❔
Fixes a low performing issue with transaction findAll query in public mode. Right now we've temporarily left two similar but not equal queries for prividium and public mode. This is until we can find the issue in prividium query and match them back together.

## Checklist


Romsters (approved):

aon:
Closed due to new workflow with `prividium-mode` branch. PRs will be done straight to that branch and a better solution will be provided for this case.

PR #459, opened 2025-06-06
feat: prividium

Romsters:


packages/worker/src/migrations/1746557256883-AddIndexesForLogsTopics.ts
```diff
@@ -0,0 +1,15 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`CREATE INDEX LogsTopic1 on public.logs ((topics[2]));`);
+    await queryRunner.query(`CREATE INDEX LogsTopic2 on public.logs ((topics[3]));`);
+    await queryRunner.query(`CREATE INDEX LogsTopic3 on public.logs ((topics[4]));`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX LogsTopic1 on public.logs;`);
+    await queryRunner.query(`DROP INDEX LogsTopic2 on public.logs;`);
+    await queryRunner.query(`DROP INDEX LogsTopic3 on public.logs;`);
```
vasyl-ivanchuk, line 13: There is no `ON` for the `DROP INDEX` statement. Should be:
```
  public async down(queryRunner: QueryRunner): Promise<void> {
    await queryRunner.query(`DROP INDEX LogsTopic1;`);
    await queryRunner.query(`DROP INDEX LogsTopic2;`);
    await queryRunner.query(`DROP INDEX LogsTopic3;`);
  }
```

packages/api/src/common/utils.ts
```diff
@@ -121,3 +122,10 @@ export const parseIntToHex = (numStr: string) => {
   }
   return "0x";
 };
+
+/**
+ * Parses the request pathname from the request object
+ */
+export const parseReqPathname = (req: Request) => {
+  return new URL(req.originalUrl, "http://localhost").pathname;
```
vasyl-ivanchuk, line 130: Can we use `req.path` instead of calling this function?

packages/app/scripts/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
Romsters, line 1: Why is this script needed? It won't work in dev mode anyway.

Romsters: Also doesn't looks like using env vars is the best way, it won't handle multiple networks, much easier to use a config file.

packages/app/src/App.vue
```diff
@@ -1,35 +1,74 @@
...
 
 useTitle(title);
 const { isReady, currentNetwork } = useContext();
 
 setup();
+
+const isAuthCheckComplete = ref(false);
+
+onMounted(async () => {
+  if (prividium.value) {
```
Romsters, line 56: Seems like this code only works if networks are set via runtime config. We should either make it work with environment config or get rid of environment config.

packages/api/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,217 @@
...
+  public async verifySignature(@Body() body: VerifySignatureDto, @Req() req: Request): Promise<boolean> {
+    if (!req.session.siwe) {
+      this.clearSession(req);
+      throw new BadRequestException({ message: "Message must be requested first" });
+    }
+
+    const siweMessage = new SiweMessage(req.session.siwe);
+    if (siweMessage.chainId !== this.configService.get("prividium.chainId")) {
+      this.clearSession(req);
+      throw new BadRequestException({ message: "Failed to verify signature" });
```
Romsters, line 112: Please also add a reason to the message, it's a common issue, will be easier to debug.

packages/api/.env.example
```diff
@@ -31,3 +32,12 @@ ETH_TOKEN_ICON_URL=https://assets.coingecko.com/coins/images/279/large/ethereum.
 ETH_TOKEN_NAME=Ether
 ETH_TOKEN_LIQUIDITY=220000000000
 ETH_TOKEN_USDPRICE=1800
+
+# Prividium
+PRIVIDIUM=false
+PRIVIDIUM_SESSION_SECRET=secret
```
Romsters, line 38: Where is it used?

packages/api/src/main.ts
```diff
@@ -50,9 +53,20 @@ async function bootstrap() {
     SwaggerModule.setup("docs", app, document);
   }
 
+  if (prividium) {
+    // Prividium config includes strict CORS configuration
+    applyPrividiumExpressConfig(app, {
+      sessionSecret: configService.get<string>("prividium.privateRpcSecret"),
```
Romsters, line 59: this is supposed to use PRIVIDIUM_SESSION_SECRET, not RPC secret

packages/api/.env.example
```diff
@@ -15,6 +15,7 @@ DISABLE_EXTERNAL_API=false
 DATABASE_STATEMENT_TIMEOUT_MS=90000
 CONTRACT_VERIFICATION_API_URL=http://127.0.0.1:3070
 NETWORK_NAME=testnet-sepolia
+APP_URL=http://localhost:3010
```
Romsters, line 18: I think it makes sense to name it PRIVIDIUM_APP_URL following the same pattern as for other prividium-specific envs. It's just irrelevant for public instances as they might be used by many apps.

packages/api/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,217 @@
...
+      domain: this.configService.get<string>("prividium.appHostname"),
+      address,
+      statement: "Sign in to the Block Explorer",
+      uri: this.configService.get("appUrl"),
+      version: "1",
+      chainId: this.configService.get("prividium.chainId"),
+      nonce: generateNonce(),
+      scheme: this.configService.get("NODE_ENV") === "production" ? "https" : "http",
+      issuedAt: new Date().toISOString(),
+      expirationTime: new Date(Date.now() + 1000 * 60 * 60).toISOString(), // 1 hour
```
Romsters, line 74: Please make it configurable via env var

packages/api/src/middlewares/auth.middleware.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    if (!req.session.siwe || !req.session.verified) {
+      req.session = null;
+      throw new UnauthorizedException({ message: "Unauthorized request" });
+    }
+
+    // Update a value in the session to reset the expiration time.
+    // Note: this is a cookie-session limitation, we can't send 'Set-Cookie'
+    // headers without modifying the session object.
+    req.session._nowInMinutes = Math.floor(Date.now() / 1000 / 60);
```
Romsters, line 24: Why is it needed to to reset the expiration time? Shouldn't it be set correctly initially?

packages/api/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    const { success, error } = await siweMessage.verify(
+      {
+        signature: body.signature,
+        nonce: siweMessage.nonce,
+        domain: this.configService.get<string>("prividium.appHostname"),
+        scheme: this.configService.get("NODE_ENV") === "production" ? "https" : "http",
+        time: new Date().toISOString(),
+      },
+      { suppressExceptions: true }
```
Romsters, line 123: A provider should be passed here as well to support EIP-1271 validation, if smart accounts for some reasons shouldn't be supported then error can be returned earlier (on message request + on the front-end).

packages/api/.env.example
```diff
@@ -31,3 +32,12 @@ ETH_TOKEN_ICON_URL=https://assets.coingecko.com/coins/images/279/large/ethereum.
...
 ETH_TOKEN_USDPRICE=1800
+
+# Prividium
+PRIVIDIUM=false
+PRIVIDIUM_SESSION_SECRET=secret
+PRIVIDIUM_PRIVATE_RPC_URL=http://private-rpc:3000
+PRIVIDIUM_PRIVATE_RPC_SECRET=secret
+PRIVIDIUM_CHAIN_ID=300
+PRIVIDIUM_SESSION_MAX_AGE=86400000 # 1 day
+PRIVIDIUM_SESSION_SAME_SITE=none
```
Romsters, line 43: I assume `PRIVIDIUM_SESSION_SAME_SITE` will always be set to `none` as API and app are hosted under different domains. It opens a possibility of a CSRF attack, but I don't see any sensitive endpoints for now. Do we plan to add any potentially sensitive endpoints for data modifications etc?

packages/app/src/components/prividium/WalletInfoModal.vue
```diff
@@ -0,0 +1,181 @@
...
+    chainName: context.currentNetwork.value.l2NetworkName,
+    l1ChainId: 0,
+    ...context.currentNetwork.value,
+  })),
+});
+
+const switchNetwork = async () => {
+  await updatePrividiumRpcUrl();
+  if (prividiumRpcUrl.value) {
+    await addNetwork(prividiumRpcUrl.value);
```
Romsters, line 102: Is it ok to add a network with a secret token in the url?

packages/app/src/composables/usePrividiumRpc.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const response = await useFetch()<{ ok: true; token: string }>(
+      `${context.currentNetwork.value.apiUrl}/auth/token`,
+      {
+        method: "POST",
+      }
+    );
+    rpcToken.value = response.token;
+    rpcUrl.value = `${context.currentNetwork.value.rpcUrl}/${response.token}`;
```
Romsters, line 29: Isn't there a better way than exposing the token in the path? Query string still not good, but better if there is no other way.

docker-compose-prividium.yaml
```diff
@@ -0,0 +1,201 @@
...
+      - LOG_LEVEL=verbose
+      - DATABASE_URL=postgres://postgres:postgres@postgres:5432/block-explorer
+      - APP_URL=http://localhost:3010
+      - PRIVIDIUM=true
+      - PRIVIDIUM_SESSION_SECRET=session_secret
+      - PRIVIDIUM_PRIVATE_RPC_URL=http://private-rpc:3000
+      - PRIVIDIUM_PRIVATE_RPC_SECRET=rpc_secret
+      - PRIVIDIUM_CHAIN_ID=270
+      - PRIVIDIUM_SESSION_MAX_AGE=86400000
+      - NODE_ENV=development # FIXME: should change http in another way
```
Romsters, line 79: What does the comment mean?

packages/api/src/address/address.controller.spec.ts
```diff
@@ -83,13 +88,15 @@ describe("AddressController", () => {
     });
 
     it("queries addresses by specified address", async () => {
-      await controller.getAddress(blockchainAddress);
+      const res = mock<Response>();
```
Romsters, line 91: Why redefining `res` instead of using `res` defined and initialized before each test?

packages/api/src/middlewares/prividium-filtering.middleware.ts
```diff
@@ -0,0 +1,148 @@
...
+          visibleBy: userAddress,
+        };
+      }
+      return;
+    }
+
+    const addressRecord = await this.addressService.findOne(reqAddress);
+    const isContract = !!(addressRecord && addressRecord.bytecode.length > 2);
+
+    if (!isContract) {
```
Romsters, line 73: What if it is a smart account?

packages/api/src/log/log.service.ts
```diff
@@ -66,4 +99,49 @@ export class LogService {
     queryBuilder.addOrderBy("log.logIndex", "ASC");
     return await queryBuilder.getMany();
   }
+
+  public async findManyByTopics({
```
Romsters, line 103: This query will be slow, existing indexes won't be used properly.

packages/api/src/log/log.service.ts
```diff
@@ -30,8 +47,24 @@ export class LogService {
     filterOptions: FilterLogsOptions = {},
     paginationOptions: IPaginationOptions
   ): Promise<Pagination<Log>> {
+    const { visibleBy, ...basicFilters } = filterOptions;
```
Romsters, line 50: This query will be slow, existing indexes won't be used properly.

packages/api/src/auth/auth.controller.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  beforeEach(() => {
+    configServiceMock = mock<ConfigService>({
+      get: jest.fn().mockImplementation((key: string) => configServiceValues[key]),
+    });
+    controller = new AuthController(configServiceMock);
+    req = mock<Request>();
+  });
+
+  describe("#getMessage", () => {
```
Romsters, line 42:
```suggestion
  describe("getMessage", () => {
```

packages/api/src/middlewares/prividium-filtering.middleware.ts
vasyl-ivanchuk, line 1: There are a number of things that concern me about `prividium-filtering.middleware`:

1. Filtering logic is all over the place. The middleware calculates the filtering params, while the actual filtering happens in the controller → service layer.
2. To attach filters to the controllers, the middleware parses the URL and relies on specific URL segments, etc. Based on this logic, it's not always clear which endpoint certain filter parameters are intended for.
3. To calculate the filter params, the middleware performs some operations that are later repeated in the controller itself. For example, it fetches the address from the DB to determine whether it's a contract — the same call is performed again in the controller.
4. Passing filter params via `res.locals` works, but it feels non-idiomatic in NestJS.

An alternative can be:
1. Write a decorator to return siwe object (pseudocode):
```
import { createParamDecorator, ExecutionContext } from '@nestjs/common';

export const User = createParamDecorator(
  (data: unknown, ctx: ExecutionContext) => {
    const request = ctx.switchToHttp().getRequest();
    return request.session.siwe;
  },
);
```
2. Change controllers to accept siwe address (pseudocode):
```
  public async getAddressLogs(
    @User('address') userAddress: string | null,
    // ... other params
  ): Promise<Pagination<LogDto>>
```
3. Move calculation of the filter params to the controllers (pseudocode):
```
if (userAddress) {
   filterParams.visibleBy = userAddress;
   // other prividium filtering logic
}
```
4. Get rid of `prividium-filtering.middleware`.

Wdyt?

packages/api/src/transaction/transaction.service.ts
```diff
@@ -54,36 +61,128 @@ export class TransactionService {
...
-      queryBuilder.leftJoin("transaction.transactionReceipt", "transactionReceipt");
-      queryBuilder.addSelect(["transactionReceipt.gasUsed", "transactionReceipt.contractAddress"]);
-      queryBuilder.leftJoin("transaction.batch", "batch");
-      queryBuilder.addSelect(["batch.commitTxHash", "batch.executeTxHash", "batch.proveTxHash"]);
-      queryBuilder.where({
-        address: filterOptions.address,
-        ...(filterOptions.receivedAt && { receivedAt: filterOptions.receivedAt }),
-        // can't add filters on transaction here due to typeOrm issue with filters on joined tables
-      });
+      const queryBuilder = this.transactionRepository.createQueryBuilder("transaction");
```
Romsters, line 64: Same as https://github.com/matter-labs/block-explorer/pull/459/files#r2136783145

packages/api/src/transfer/transfer.service.ts
```diff
@@ -52,12 +53,34 @@ export class TransferService {
     filterOptions: FilterTransfersOptions = {},
     paginationOptions: IPaginationOptions
   ): Promise<Pagination<Transfer>> {
-    if (filterOptions.address) {
+    const { visibleBy, ...basicOptions } = filterOptions;
+
+    if (visibleBy) {
```
Romsters, line 58: Same as https://github.com/matter-labs/block-explorer/pull/459/files#r2136783145

packages/app/scripts/entrypoint.sh
aon: Changed it in cd8b202521df4d9dd19660b2f44d2c28ebc0a284

packages/api/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,217 @@
...
+      ok: z.literal(true),
+      token: z.string().min(1),
+    });
+
+    const result = schema.safeParse(response);
+    if (!result.success) {
+      // Left as 500 since now everyone is able to generate RPC tokens.
+      // In the future, this must be changed to handle unauthorized responses, once the
+      // Private RPC is adapted.
+      throw new InternalServerErrorException();
```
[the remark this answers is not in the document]
aon: Added message in ec90436e1c696154fd54861c567cf2bc0bb6b736

packages/api/src/auth/auth.controller.ts
aon: Fixed in ec90436e1c696154fd54861c567cf2bc0bb6b736

packages/api/src/main.ts
aon: Fixed in f204a4cac17ccd341e7892fc46262e5434f54a55

packages/api/.env.example
aon: Fixed in f204a4cac17ccd341e7892fc46262e5434f54a55, it was previously unused

packages/api/src/auth/auth.controller.ts
aon: Agreed, changed .env.example to use 10 minutes. Added in f204a4cac17ccd341e7892fc46262e5434f54a55

packages/api/.env.example
aon: API and app might be hosted under a same domain with different subdomains. Adding this same site configuration allows for more flexibility. I think so far we don't have plans to adding sensitive endpoints for data modifications in block explorer.

packages/app/src/composables/useContractInteraction.ts
```diff
@@ -90,11 +112,20 @@ export default (context = useContext()) => {
...
+      if (context.currentNetwork.value.prividium) {
+        signer = new Provider(context.currentNetwork.value.rpcUrl);
+        if (walletAddress.value !== null) {
+          signer = await getL2Signer();
+        }
+      } else {
+        signer = new Provider(context.currentNetwork.value.rpcUrl);
+        if (walletAddress.value !== null) {
+          signer = await getL2Signer();
+        }
```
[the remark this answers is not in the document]
aon: Fixed in c99fa106f4a8e8f41fef0fd17251a2374eab526e

packages/api/src/auth/auth.controller.ts
aon: Thanks for raising this. So far we only support Metamask in the block explorer app, so we don't need EIP-1271 validation. We will take this consideration in the future.

packages/api/src/address/address.controller.spec.ts
```diff
@@ -83,13 +88,15 @@ describe("AddressController", () => {
...
+      await controller.getAddress(blockchainAddress, res);
       expect(serviceMock.findOne).toHaveBeenCalledTimes(1);
       expect(serviceMock.findOne).toHaveBeenCalledWith(blockchainAddress);
     });
 
     it("queries address balances", async () => {
-      await controller.getAddress(blockchainAddress);
+      const res = mock<Response>();
```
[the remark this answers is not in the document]
aon: Fixed in 73309f66a7966a2b99b9af717a3c42e59f59c49c

packages/api/src/app.module.ts
```diff
@@ -53,7 +55,55 @@ import config from "./config";
...
+
+  constructor(@Inject(PRIVIDIUM_TOKEN) moduleConfig: AppModuleConfig, allConfigs: ConfigService) {
+    this.prividium = moduleConfig.prividium;
+
+    if (this.prividium) {
+      const schema = z.object({
+        privateRpcUrl: z.string().url(),
+        privateRpcSecret: z.string().min(1),
+      });
+      const result = schema.safeParse(allConfigs.get("prividium"));
```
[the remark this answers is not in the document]
aon: Removed in d39829950d1e2944924052d4f4612a50c04b4b52

packages/api/src/middlewares/prividium-filtering.middleware.ts
aon: Agreed with the approach, PR: https://github.com/matter-labs/block-explorer/pull/473

packages/app/src/locales/en.json
```diff
@@ -1,816 +1,829 @@
 {
-    "general": {
```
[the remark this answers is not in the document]
aon: Looks like indentation changed from 4 to 2 spaces, fixed.

packages/api/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,217 @@
...
+    const siweMessage = new SiweMessage(req.session.siwe);
+    if (siweMessage.chainId !== this.configService.get("prividium.chainId")) {
+      this.clearSession(req);
+      throw new BadRequestException({ message: "Failed to verify signature" });
+    }
+
+    const { success, error } = await siweMessage.verify(
+      {
+        signature: body.signature,
+        nonce: siweMessage.nonce,
```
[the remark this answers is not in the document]
Romsters: yeah, but if somebody steals the nonce somehow then the request can be replayed. Probably this is unlikely to happen, but checking if it is already used would prevent this.

packages/app/src/locales/en.json
```diff
@@ -813,5 +813,18 @@
...
+        "addressLabel": "Address",
+        "balanceLabel": "Balance",
+        "disconnectButton": "Disconnect",
+        "balanceUnavailable": "Balance unavailable",
+        "loadingBalance": "Loading balance...",
+        "noEthBalance": "ETH balance not found",
+        "wrongNetwork": "Your wallet activated an incorrect network",
+        "switchNetwork": "Switch to Prividium"
+    },
+    "Login": "Login"
```
Romsters, line 829: where is it used?

packages/app/tests/e2e/src/support/common-hooks.ts
```diff
@@ -1,48 +1,95 @@
...
-  return "skipped" as any;
+  return Status.SKIPPED;
 });
 
-Before(async function (this: ICustomWorld, { pickle }: ITestCaseHookParameter) {
+Before({ tags: "not @prividium" }, async function (this: ICustomWorld, { pickle }: ITestCaseHookParameter) {
   this.testName = pickle.name.replace(/\W/g, "-");
-  this.context = await browser.newContext({
+
+  if (!legacyBrowser) {
```
Romsters, line 25: why is it called legacy?

packages/app/src/locales/en.json
aon: Deleted, it was unused

aon: Removed

packages/app/src/components/ConnectMetamaskButton.vue
```diff
@@ -1,46 +1,67 @@
 <template>
   <div class="metamask-button" :class="{ disabled: buttonDisabled }">
```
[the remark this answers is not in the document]
aon: Fixed in ca6073c26dc1667f28423d46629f71c64df7391f. Connect metamask button was split into two. When on prividium mode, connect button doesn't show up when attempting to read/write from a contract as this functionality is already covered in the header button.

packages/app/src/views/NotAuthorizedView.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <div class="fixed inset-0 flex flex-col items-center justify-center bg-[#11142B] text-white">
+    <div class="flex flex-col items-center text-center max-w-md mx-auto px-4">
+      <!-- Logo -->
+      <div class="mb-10">
+        <img src="/images/zksync-light.svg" class="w-[233px] h-[48px]" alt="ZKsync Logo" />
+      </div>
+
+      <!-- Heading -->
+      <h1 class="text-[30px] leading-[36px] font-bold mb-2">You are not authorized</h1>
```
Romsters, line 10: Please move copies to the en.json file

packages/app/src/views/LoginView.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <div class="fixed inset-0 flex flex-col items-center justify-center bg-[#11142B]">
+    <div class="p-8 rounded-lg max-w-xxl w-full flex flex-col items-center">
+      <div class="flex justify-center mb-10">
+        <img src="/images/zksync-light.svg" class="w-[233px] h-[48px]" />
+      </div>
+      <h1 class="text-[30px] leading-[36px] font-bold tracking-[0%] mb-2 text-center text-white">
+        Private Explorer Access
```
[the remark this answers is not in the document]
aon: Fixed in 5ca505a95159d5af80428ebefef95e3f32333c68

vasyl-ivanchuk (approved):

PR #460, opened 2025-06-06
fix: external tokens data for any chain

bxpana:
# What ❔

- Fix CoingeckoTokenOffChainDataProvider so external tokens data can be fetched for any chain.

## Why ❔

Allows coingecko integration to retrieve prices for configured chain instead of hardcoded zksync.

## Checklist


packages/app/src/configs/local.config.json
vasyl-ivanchuk, line 1: This needs to be rolled back to the local environment settings.

packages/worker/src/token/tokenOffChainData/providers/coingecko/coingeckoTokenOffChainDataProvider.ts
```diff
@@ -49,49 +50,52 @@ export class CoingeckoTokenOffChainDataProvider implements TokenOffChainDataProv
...
-        locale: "en",
-      },
+    const tokenIds = supportedTokens.map((token) => token.id);
+    const tokensMarkedData = await this.makeApiRequest<ITokenMarketDataProviderResponse[]>("/coins/markets", {
+      vs_currency: "usd",
+      ids: tokenIds.join(","),
+      per_page: tokenIds.length.toString(),
+      page: "1",
+      locale: "en",
     });
```
vasyl-ivanchuk, line 80: 1. The previous implementation respected the API response's max page size of 250, which is why it requested information about tokens in batches of up to 250 token ids each. This PR removes this logic and sends a single request, which may not be sufficient, especially on Era mainnet.

2. It also used `makeApiRequestRetryable` function, which handles `HTTP 429 Too Many Requests` and implements retries with gradually increasing timeouts. The new functionality no longer uses this approach, which makes the processing less reliable.

packages/app/src/configs/local.config.json
vasyl-ivanchuk: `packages/app/src/configs/local.config.json` contains Sophon network settings which is incorrect.

packages/worker/src/token/tokenOffChainData/providers/coingecko/coingeckoTokenOffChainDataProvider.ts
```diff
@@ -35,12 +35,14 @@ export class CoingeckoTokenOffChainDataProvider implements TokenOffChainDataProv
...
   private readonly apiKey: string;
   private readonly apiUrl: string;
+  private readonly platformId: string;
 
   constructor(configService: ConfigService, private readonly httpService: HttpService) {
     this.logger = new Logger(CoingeckoTokenOffChainDataProvider.name);
     this.isProPlan = configService.get<boolean>("tokens.coingecko.isProPlan");
     this.apiKey = configService.get<string>("tokens.coingecko.apiKey");
     this.apiUrl = this.isProPlan ? "https://pro-api.coingecko.com/api/v3" : "https://api.coingecko.com/api/v3";
+    this.platformId = configService.get<string>("COINGECKO_PLATFORM_ID", "zksync"); // Default to zksync for backward compatibility
```
vasyl-ivanchuk, line 45: To keep things consistent, let's move usage of `COINGECKO_PLATFORM_ID` to the [config module](https://github.com/matter-labs/block-explorer/blob/main/packages/worker/src/config.ts#L96) so it's used the same way as other env variables:

```
      coingecko: {
        isProPlan: COINGECKO_IS_PRO_PLAN === "true",
        apiKey: COINGECKO_API_KEY,
        platformId: COINGECKO_PLATFORM_ID || "zksync",
      },
```
and then use it like this:
```
configService.get<string>("tokens.coingecko.platformId")
```

Romsters (approved):

Issue #454, opened 2025-05-30
Txs are not decoded for proxy contracts


bxpana:
#### 📝 Description

Same as the parent issue, but function and params for the Transactions page in the "Method" column and within the tx information are not decoded. 

This might be due to the fact that the contracts are "auto-verified"

#### 🔄 Reproduction Steps

[Sophon example](https://explorer.testnet.sophon.xyz/address/0x3951eDF6D554FcB0327b4D5218Dd1B8BD90B456e)
[Lens example](https://explorer.lens.xyz/address/0xc6d57ee750ef2ee017a9e985a0c4198bed16a802#transactions)

#### 🤔 Expected Behavior

To see the decoded input data and params in the tx "Input data" and in the "Method" column of the Transactions page. 

#### 😯 Current Behavior

Input data is not decoded and function names do not show in "Method" column

#### 🖥️ Environment

This might be due to the fact that the contracts are "auto-verified"

#### 📋 Additional Context

<img width="1293" alt="Image" src="https://github.com/user-attachments/assets/936bcac4-7b86-408d-9c74-fcd14f75f891" />

<img width="1267" alt="Image" src="https://github.com/user-attachments/assets/9139c232-0cc1-4f65-b36c-cfcceae30808" />

#### 📎 Log Output

```
Paste any relevant log output here.
```

PR #461, opened 2025-06-06
feat: decode proxy contracts

bxpana:
# What ❔

- Auto-verified contracts now properly decode input data, functions and parameters just like manually verified contracts
- Add a new `partialVerification` computed property to `CompilationInfo.vue` to indicate if the contract is only partially verified
- Add a new `autoVerified` computed property to `CompilationInfo.vue` to indicate if the contract is auto-verified
- Add a new `PARTIAL_VERIFICATION_DETAILS_URL` constant to `CompilationInfo.vue` to link to the documentation for partial verification

## Why ❔

Allows contracts that were auto-verified to have the input data decoded without having to manually verify them again. Should resolve https://github.com/matter-labs/block-explorer/issues/454

## Checklist

### DIFF
--- a/packages/app/src/composables/useTransactionData.ts
+++ b/packages/app/src/composables/useTransactionData.ts
@@ -82,12 +82,6 @@ export default (context = useContext()) => {
       }
 
       const contractAbi = ABICollection.value[transactionData.contractAddress];
-      const proxyInfo = await getContractProxyInfo(transactionData.contractAddress);
-
-      // Check for contract verification
-      if (!contractAbi && !proxyInfo?.implementation.verificationInfo) {
-        throw new Error("contract_not_verified");
-      }
 
       // Try to decode with contract's own ABI first if available
       let method: TransactionData["method"] | undefined;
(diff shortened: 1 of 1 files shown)


packages/app/src/components/contract/CompilationInfo.vue
```diff
@@ -136,6 +136,10 @@ const autoVerified = computed(
 // If there are any problems with the verification, the contract is only partially verified.
 const partialVerification = computed(() => {
   const problems = props.contract.verificationInfo?.verificationProblems;
+  // For auto-verified contracts, we don't consider verification problems since we have the ABI
+  if (autoVerified.value) {
```
vasyl-ivanchuk, line 140: It's possible that a contract was partially auto-verified. This happens when:
- The contract itself was not verified.
- There is only a partial match with other verified contracts.

In such cases, we show two messages:
image

Example of such contract: https://sepolia.explorer.zksync.io/address/0xad8EB0B86dafcfF7B36e47156Eef8920f7C6a9f6#contract

With this change, in such cases only one message will be shown that says: _The contract wasn't manually verified, but its bytecode matches the bytecode of the following contract._

image

Without the second message, it's a bit misleading, as it's not clear that the match is partial. I don't mind showing just one message in such a case, but then I think we need to specify that it's not a full match. For instance:
_The contract wasn't manually verified, but its bytecode partially matches the bytecode of the following contract ... (the contract metadata hash and/or constructor arguments do not match)._

Wdyt?

packages/app/src/composables/useTransactionData.ts
```diff
@@ -72,30 +72,45 @@ export default (context = useContext()) => {
...
       }
-      let method = abi ? decodeDataWithABI(transactionData, abi) : undefined;
-      if (!method) {
-        const proxyInfo = await getContractProxyInfo(transactionData.contractAddress);
-        method = proxyInfo?.implementation.verificationInfo
-          ? decodeDataWithABI(transactionData, proxyInfo.implementation.verificationInfo.artifacts.abi)
-          : undefined;
+
+      const contractAbi = ABICollection.value[transactionData.contractAddress];
+      const proxyInfo = await getContractProxyInfo(transactionData.contractAddress);
```
vasyl-ivanchuk, line 85: `getContractProxyInfo` performs multiple calls under the hood. Can we call it only when `contractAbi` doesn't work?

packages/app/src/components/contract/CompilationInfo.vue
bxpana: Went with keeping the first message when a contract is partially auto-verified. I like the idea of a single message, but didn't want to change too much right now.

vasyl-ivanchuk:
LGTM!

Issue #406, opened 2025-04-15
Use `debug_traceTransaction` to decode revert reasons


bxpana:
#### 📝 Description

Currently the explorer uses `zks_getTransactionDetails` to get transaction details and returns `Bootloader-based tx failed` instead of the revert message

#### 🤔 Rationale

Provides better UI/UX for devs and users to understand why their transaction failed

#### 📋 Additional Context

Etherscan versions of the explorer do query traces and show proper error messages
- [ZKsync Sophon testnet Explorer](https://explorer.testnet.sophon.xyz/tx/0xdacd73f201c4227fdcf18a4fc74c6bfca9538b3677250950320e3d6bb1ab4250)
- [Etherscan Sophon testnet Explorer](https://testnet.sophscan.xyz/tx/0xdacd73f201c4227fdcf18a4fc74c6bfca9538b3677250950320e3d6bb1ab4250)

PR #466, opened 2025-06-13
feat: add error message to revert reason

bxpana:
# What ❔

- Modified the transaction reason display logic in packages/app/src/components/transactions/infoTable/GeneralInfo.vue to show error message and revert reason message when the transaction is reverted.
- Added CSS styling to ensure proper spacing between the error and revert reason messages when both are displayed.

- Can test on Era mainnet tx: `0x51640769e14c9b8ac7270a3deaf1fbb8f0c5c663c3c967527cd484539461c191`
## Why ❔

Shows revert reason and error message on block explorer frontend for easier debugging. Resolves https://github.com/matter-labs/block-explorer/issues/406

## Checklist


cytadela8 (approved):

packages/app/src/components/transactions/infoTable/GeneralInfo.vue
```diff
@@ -40,7 +40,16 @@
           </InfoTooltip>
         </TableBodyColumn>
         <TableBodyColumn class="transaction-table-value transaction-reason-value">
-          {{ transaction.error || transaction.revertReason || "" }}
+          <div v-if="transaction?.error && transaction?.revertReason">
```
Romsters, line 43: This is not right, means it is only shown if both error and revertReason are present which is not always the case

packages/app/src/components/transactions/infoTable/GeneralInfo.vue
```diff
@@ -353,6 +366,9 @@ const gasUsedPercent = computed(() => {
   .transaction-status-value {
     @apply py-2;
   }
+  .transaction-error-value {
```
vasyl-ivanchuk, line 369: `.transaction-error-value` and `.transaction-reason-value` are identical. To avoid duplication, let's leave just `.transaction-error-value` and use for both error and error reason.

PR #469, opened 2025-06-17
feat: Prividium whitelisting for login functionality

Jrigada:
- Added a whitelist check for wallet addresses in the authentication process
- Updated the login handling components to redirect users to a not-authorized page on 403 errors.

# What ❔
- Added a whitelist check for wallet addresses in the authentication process
- Updated the login handling components to redirect users to a not-authorized page on 403 errors.

## Checklist

### DIFF
--- a/packages/api/src/auth/auth.controller.ts
+++ b/packages/api/src/auth/auth.controller.ts
@@ -9,6 +9,7 @@ import {
   UnprocessableEntityException,
   BadRequestException,
   InternalServerErrorException,
+  ForbiddenException,
 } from "@nestjs/common";
 import {
   ApiTags,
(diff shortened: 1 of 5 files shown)


packages/api/src/auth/auth.controller.ts
```diff
@@ -194,6 +201,27 @@ export class AuthController {
...
+      // For other errors, we'll deny access by default.
+      return false;
+    }
+
+    const data = await response.json();
+    const schema = z.object({
+      authorized: z.boolean(),
+    });
+    const validation = schema.safeParse(data);
+    return validation.success && validation.data.authorized;
```
aon, line 222: Schema can be defined outside. If schema is not fulfilled, it's a 500.

packages/app/src/types.ts
```diff
@@ -69,3 +69,14 @@ export enum TimeFormat {
...
+export type HttpError = {
+  response?: {
+    status?: number;
+  };
+  message?: string;
+};
+
+export const isHttpError = (error: unknown): error is HttpError => {
+  return typeof error === "object" && error !== null;
+};
```
aon, line 82: I'm not sure this completely asserts error is of type http error. Isn't there a `FetchError` type to derive from?

aon (approved):

calvogenerico (approved):

PR #472, opened 2025-06-23
feat: use tx traces for contract deployments

vasyl-ivanchuk:
# What ❔

Use transaction traces to get information about the deployed contracts.

## Why ❔

Once `ContractDeployed` events are removed, the explorer won’t be able to detect contract deployments, so we need to use transaction traces for this purpose.

## Checklist


packages/data-fetcher/src/transaction/transactionTraces.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  if (transactionTrace) {
+    if (transactionTrace.type === "create") {
+      extractedData.contractAddresses.push({
+        address: transactionTrace.to,
+        blockNumber: transactionReceipt.blockNumber,
+        transactionHash: transactionReceipt.hash,
+        creatorAddress: transactionReceipt.from,
+        logIndex: extractedData.contractAddresses.length + 1,
```
Romsters, line 43: This is just to keep something in logIndex, right?

Romsters (approved):

PR #499, opened 2025-07-31
fix: settlement layer icon when txs are processed by different layers

vasyl-ivanchuk:
# What ❔

Currently, the settlement layer icon for batches and transactions is determined solely based on the `commitChainId` field. As a result, an incorrect settlement layer icon may be shown if the commit, prove and execute transactions are processed by different chains. For example:
https://explorer.zksync.io/batch/501263, where the commit and execute transactions are processed by Ethereum and ZKsync GW, respectively.

image
image 

To fix this:
1. The API has been updated to return `proveChainId` and `executeChainId` for batch list items.
2. When displaying the settlement layer icon, the appropriate chain ID is used, not just `commitChainId`.

After the fix:
image
image

## Why ❔

To show a proper settlement layer icon.

## Checklist

### DIFF
--- a/packages/app/src/components/batches/Table.vue
+++ b/packages/app/src/components/batches/Table.vue
@@ -97,7 +97,11 @@ function getBadgeIconByStatus(batch: BatchListItem) {
   if (batch.status === "sealed") {
     return ZkSyncIcon;
   }
-  return isGatewaySettlementChain(batch.commitChainId) ? GatewayIcon : EthereumIcon;
+  debugger;
+  // Check chain IDs beginning with the most recent status and moving backward, in case transactions are processed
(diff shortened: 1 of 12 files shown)


packages/app/src/components/batches/Table.vue
```diff
@@ -97,7 +97,11 @@ function getBadgeIconByStatus(batch: BatchListItem) {
   if (batch.status === "sealed") {
     return ZkSyncIcon;
   }
-  return isGatewaySettlementChain(batch.commitChainId) ? GatewayIcon : EthereumIcon;
+  debugger;
```
Romsters, line 100: Debugger

Romsters (dismissed):

Romsters (approved):

PR #507, opened 2025-08-12
feat: add login with prividium

aon:
# What ❔

This PR adds login with User Panel. User Panel is an app in Prividium that is in charge of managing the user's account and allowing third-party services to login. Right now, it's only enabled for IdP (Okta).

## Why ❔

To support login with Prividium User Panel.

## Checklist

### DIFF
--- a/packages/app/src/composables/useLogin.ts
+++ b/packages/app/src/composables/useLogin.ts
@@ -72,10 +72,6 @@ export default (context: Context, _logger = defaultLogger): UseLogin => {
     try {
       state.isLoginPending = true;
       const auth = getPrividiumAuth();
-      if (!auth) {
-        throw new Error("Prividium authentication is not configured");
-      }
-
       const result = auth.handleCallback();
 
       if (result && result.token) {
@@ -96,7 +92,7 @@ export default (context: Context, _logger = defaultLogger): UseLogin => {
 
   const logout = async () => {
     const auth = getPrividiumAuth();
-    auth?.clearToken();
+    auth.clearToken();
 
     try {
       await FetchInstance.api(context)("/auth/logout", {
--- a/packages/api/src/common/utils.spec.ts
+++ b/packages/api/src/common/utils.spec.ts
@@ -419,35 +419,35 @@ describe("utils", () => {
   });
 
   describe("parseReqPathname", () => {
-    it("should return the pathname for a simple path", () => {
+    it("returns the pathname for a simple path", () => {
       const req = {
         originalUrl: "/test",
       } as any;
       expect(parseReqPathname(req)).toBe("/test");
     });
 
-    it("should return the pathname when query parameters are present", () => {
+    it("returns the pathname when query parameters are present", () => {
       const req = {
         originalUrl: "/test?foo=bar&baz=qux",
       } as any;
       expect(parseReqPathname(req)).toBe("/test");
     });
 
-    it("should return the pathname for a complex path", () => {
+    it("returns the pathname for a complex path", () => {
       const req = {
         originalUrl: "/api/v1/items/123",
       } as any;
       expect(parseReqPathname(req)).toBe("/api/v1/items/123");
     });
 
-    it("should return '/' for the root path", () => {
+    it("returns '/' for the root path", () => {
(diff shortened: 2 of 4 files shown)


packages/api/src/auth/auth.controller.ts
vasyl-ivanchuk, line 1: I see that `siwe` is no longer used. The only file where it’s still referenced is `siwe-message-tools`, which itself is not used. Let’s remove it and delete the siwe dependency from package.json.

packages/api/src/config/index.ts
vasyl-ivanchuk, line 1: Prividium `siweExpirationTime` config is not used anymore. It, together with its dependencies (such as validation error messages), has to be removed from the config module and the env.example file.

.github/workflows/prividium-e2e.yml
```diff
@@ -3,10 +3,10 @@ name: Prividium E2E Tests
...
   pull_request:
     paths:
-      - "packages/app/**"
-      - "packages/api/**"
-      - ".github/workflows/prividium-e2e.yml"
-      - "docker-compose-prividium.yaml"
+      - 'packages/app/**'
+      - 'packages/api/**'
+      - '.github/workflows/prividium-e2e.yml'
+      - 'docker-compose-prividium.yaml'
```
vasyl-ivanchuk, line 9: Do we need to update `docker-compose-prividium.yaml` with the new config? It uses old env variables such as `PRIVIDIUM_PRIVATE_RPC_URL` and `PRIVIDIUM_PRIVATE_RPC_SECRET` and doesn't set `PRIVIDIUM_PERMISSIONS_API_URL`.

packages/api/src/middlewares/auth.middleware.ts
```diff
@@ -13,7 +14,12 @@ export class AuthMiddleware implements NestMiddleware {
...
-    if (!req.session.siwe || !req.session.verified) {
+    if (!req.session.address || !req.session.token) {
+      req.session = null;
+      throw new UnauthorizedException({ message: "Unauthorized request" });
+    }
+
+    if (!isAddress(req.session.address)) {
       req.session = null;
       throw new UnauthorizedException({ message: "Unauthorized request" });
     }
```
vasyl-ivanchuk, line 25:
```suggestion
    if (!req.session.address || !req.session.token || !isAddress(req.session.address)) {
      req.session = null;
      throw new UnauthorizedException({ message: "Unauthorized request" });
    }
```

packages/api/src/auth/auth.controller.spec.ts
```diff
@@ -49,449 +30,78 @@ describe("AuthController", () => {
...
 
-    it("throws when there is no message", async () => {
-      await expect(() => controller.verifySignature(body, req)).rejects.toThrow(BadRequestException);
-    });
-
-    it("validates for http in development", async () => {
-      // Mock the whitelist check to succeed
-      fetchSpy.mockResolvedValue({
-        ok: true,
+    it("should login successfully with valid token", async () => {
```
vasyl-ivanchuk, line 46: Let's stick to the existing wording style for test cases:
`should login ...` -> `logins`
`should throw ...` -> `throws`

This way, the test cases read like a specification. Same in other places.

packages/app/src/composables/useLogin.ts
```diff
@@ -16,26 +16,37 @@ type UseLogin = ToRefs<LoginState> & {
...
-      l1ChainId: 0,
-      ...context.currentNetwork.value,
-    })),
-  });
+  const router = useRouter();
+
+  const getPrividiumAuth = () => {
+    if (!context.currentNetwork.value.prividium) {
+      return null;
+    }
```
vasyl-ivanchuk, line 34: Does it make sense to throw an Exception here instead of returning a `null`?
```
    if (!context.currentNetwork.value.prividium) {
      throw new Error("Prividium authentication is not configured");
    }
```
It's thrown up the callstack anyway. This way it will be in one place.

vasyl-ivanchuk:
LGTM!

vasyl-ivanchuk (approved):

PR #582, opened 2026-02-13
feat: apply event log filtering rules from prividium api

jodator:
# What ❔

* Implement log filtering logic based on event permission rules fetched from Prividium API
* Introduces rule‑based log visibility with a toggleable policy and permissions API integration.
* Updates controllers to pass visibility context
* Updated routes:
  * `GET /address/:address/logs`
  * `GET /transactions/:transactionHash/logs`
  * `GET /transactions/:transactionHash` (fetches logs to determine transaction visibility)
* removed restrictions from the UI - now any user can access logs on contract and transaction pages

## Why ❔

* Implement the same level of event log filtering as it is enforced by the Prividium RPC endpoint
* Group prividium-related code for better auditability of restrictions

Closes matter-labs/zksync-prividium#700
Closes matter-labs/zksync-prividium#131

Part of matter-labs/zksync-prividium#704

## Checklist

- [ ] ~Documentation comments have been added / updated.~


packages/api/src/log/log.tokens.ts
```diff
@@ -0,0 +1,4 @@
+export const LOG_VISIBILITY_POLICY = Symbol("LOG_VISIBILITY_POLICY");
```
Romsters: `log.tokens.ts` filename is confusing. Please use `log.utils.ts` or just declare it in `common/constants` or `common/utils`.

packages/api/src/prividium/prividium-provider.factory.ts
```diff
@@ -0,0 +1,15 @@
+import { FactoryProvider } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+
+export function makePrividiumToggleProvider<T>(
```
Romsters:
```suggestion
export function createPrividiumConditionalProvider<T>(
```

packages/api/src/prividium/policies/log-visibility.policy.ts
```diff
@@ -0,0 +1,103 @@
...
+    eventPermissionUserAddress?: string
+  ) {
+    if (eventPermissionRules.length === 0) {
+      qb.andWhere("FALSE");
+      return;
+    }
+
+    qb.andWhere(
+      new Brackets((outer) => {
+        eventPermissionRules.forEach((rule, idx) => {
```
Romsters: There is no way it can work with current indexes with large amount of data

packages/api/src/prividium/policies/log-visibility.policy.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+@Injectable()
+export class RuleBasedLogVisibilityPolicy implements LogVisibilityPolicy {
+  constructor(private readonly rulesService: PrividiumRulesService) {}
+
+  async apply(qb: SelectQueryBuilder<Log>, visibility?: VisibilityContext): Promise<void> {
+    const user = visibility?.user;
+    // Deny if no user or missing token
+    if (!user || !user.token) {
```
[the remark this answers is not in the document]
jodator: The main case for using explicit "Noop" vs "Prividium" is to enforce rule checking when the prividium is configured. In prividium mode I wanted to force using this policy even if the `visibility` is not passed (by mistake for instance) - otherwise misuse in code will leak by default. I'd rather not rely on param passed to detect if that's Prividium or not.

Util or helper is also OK at this stage - one use case. I'll propose something simpler for now but having prividium rules in mind.

PR #613, opened 2026-06-01
fix(data-fetcher): authenticate bridge event emitter before indexing transfers

vladb-ai:
## Problem

The bridge transfer handlers in `data-fetcher` select work purely by event **topic** and their `matches()` returned `true` unconditionally, never checking the log **emitter**. As a result any unprivileged L2 contract can emit ABI-shaped bridge logs and have them indexed as **real bridge deposits/withdrawals** with attacker-chosen `from` / `to` / `amount` / `token`:

- `DepositFinalizedAssetRouter` / `WithdrawalInitiatedAssetRouter` (Asset Router path)
- `FinalizeDeposit` / `WithdrawalInitiated` (legacy shared-bridge path)

These fabricated transfers surface in the explorer UI and the public API (e.g. on a victim address's `/transfers`), i.e. misrepresentation of transaction data. No fund safety impact; data-integrity only.

## Fix

**Asset Router handlers** — require the emitter to be the L2 Asset Router system contract (`L2_ASSET_ROUTER_ADDRESS`, `0x…010003`). This is deployed at the same fixed address on Era and every ZK chain, so the check is chain-agnostic with no config.

**Legacy shared-bridge handlers** — legacy events have **no single authoritative emitter** (the canonical shared bridge *plus* deployment-specific custom token bridges, e.g. Lido wstETH). They are authenticated against a trusted set built from:
1. `zks_getBridgeContracts` (canonical bridges; cached), and
2. a configurable allowlist via the new `TRUSTED_LEGACY_BRIDGE_ADDRESSES` env var (comma-separated).

Chains that don't implement `zks_getBridgeContracts` (e.g. ZKsync OS) fall back to the allowlist; they emit no legacy events.

## Verification

- **Unit tests:** full `data-fetcher` suite green (400 tests; new accept/reject cases for both handler families + `getTrustedLegacyBridgeAddresses`). Lint clean.
- **Cross-chain (live RPC):** every Asset Router deposit/withdrawal on **Era mainnet** (443+ samples) and the deposit on **ZKsync OS mainnet** is emitted solely by `0x…010003`; the patched handler accepts those and rejects other emitters.
- **Legacy (real Era log):** against the real Lido wstETH withdrawal — canonical-only set drops it (the trade-off), `TRUSTED_LEGACY_BRIDGE_ADDRESSES` re-includes it, and an attacker emitter is rejected.
- **End-to-end:** rebuilt data-fetcher rejects real on-chain attacker-emitted bridge logs while still parsing router/bridge-attributed ones.

## Operator note

To keep indexing legitimate **custom** legacy bridges, set `TRUSTED_LEGACY_BRIDGE_ADDRESSES`. For Era mainnet this includes at least the Lido wstETH bridge `0xe1d6a50e7101c8f8db77352897ee3f1ac53f782b`. Recommend scanning historical legacy-topic emitters per network to seed this list before rollout so no legitimate custom-bridge transfers are dropped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/data-fetcher/.env.example
+++ b/packages/data-fetcher/.env.example
@@ -19,3 +19,11 @@ RPC_BATCH_STALL_TIME_MS=0
 MAX_BLOCKS_BATCH_SIZE=20
 
 RPC_HEALTH_CHECK_TIMEOUT_MS=20_000
+
+# Comma-separated list of L2 addresses (besides the canonical bridges reported by
+# zks_getBridgeContracts) whose legacy FinalizeDeposit / WithdrawalInitiated events should be
+# indexed as real bridge transfers, e.g. third-party custom token bridges. Deployment-specific:
+# emitters not in this list and not a canonical bridge are skipped and counted via the
+# skipped_untrusted_legacy_bridge_logs_total metric. For ZKsync Era mainnet this should include at
+# least the Lido wstETH bridge.
+# TRUSTED_LEGACY_BRIDGE_ADDRESSES=0xe1d6a50e7101c8f8db77352897ee3f1ac53f782b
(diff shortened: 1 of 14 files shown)


Romsters:
Closing in favour of https://github.com/matter-labs/block-explorer/pull/619 and https://github.com/matter-labs/block-explorer/pull/620

PR #618, opened 2026-06-03
fix(prividium): route page-load failures to a privacy-preserving not-found page

tomimor:
# What ❔

- In Prividium mode, address/block/token/transaction page loads that fail with a 4xx now land on the existing `/not-found` page with the copy "Nothing to show. This page may not exist, or it may not be visible to you." instead of a red "Something went wrong" card or the search-style default copy.

image

## Why ❔

- Non-admins navigating to another user's address or blocked pages saw a generic "Something went wrong" that read like a bug, not an expected outcome.

Before: 
image

Related: matter-labs/zksync-prividium#172

## Note

- Contact support being hardcoded is currently an existing bug in the not-found page. As this extends to the footer as well, I will address this in a separate PR. 

## Checklist


packages/app/src/composables/useAddress.ts
```diff
@@ -192,7 +192,9 @@ export default (context = useContext()) => {
       }
     } catch (error: unknown) {
       item.value = null;
-      isRequestFailed.value = true;
+      if (!(error instanceof FetchError)) {
+        isRequestFailed.value = true;
+      }
```
aon: I don't think we should skip all `FetchError` types. We want to keep current behaviour for 400/500 type of errors. If we want to capture particularly the `401` error, then we should add a middleware for this purpose.

packages/app/src/composables/useBlock.ts
```diff
@@ -42,7 +42,7 @@ export default (context = useContext()) => {
       blockItem.value = await FetchInstance.api(context)(`/blocks/${id}`);
     } catch (error: unknown) {
       blockItem.value = null;
-      if (!(error instanceof FetchError) || error.response?.status !== 404) {
```
calvogenerico, line 45: Why was the check for 404 removed?

packages/app/src/locales/en.json
```diff
@@ -805,6 +805,8 @@
     "notFound": {
         "title": "Oops, we can’t find anything",
         "description": "Sorry we can’t find anything on your search result, please try one more time via search bar below",
+        "prividiumTitle": "Nothing to show",
+        "prividiumDescription": "This page may not exist, or it may not be visible to you.",
```
calvogenerico:
```suggestion
        "prividiumDescription": "This page may not exist, or you might not have permissions to see it.",
```

packages/app/src/composables/useAddress.ts
tomimor: Fixed in [2eba948](https://github.com/matter-labs/block-explorer/pull/618/commits/2eba94870700e4da96251c79e1f3a5b06ec9e179)

packages/app/src/composables/useBlock.ts
tomimor: Fixed in [2eba948](https://github.com/matter-labs/block-explorer/pull/618/commits/2eba94870700e4da96251c79e1f3a5b06ec9e179)

packages/app/src/locales/en.json
tomimor: Fixed in [2eba948](https://github.com/matter-labs/block-explorer/pull/618/commits/2eba94870700e4da96251c79e1f3a5b06ec9e179)

Romsters (approved):

PR #621, opened 2026-06-08
feat(prividium): serve Swagger /docs publicly

tomimor:
# What ❔

Removes the auth gate (`applySwaggerAuthMiddleware`) that was returning 401 on `/docs` in Prividium mode. The `/docs` route is now reachable without a session, matching the Permissions API's public-docs behavior.

## Why ❔

- Block-explorer `/docs` returned 401 in Prividium mode because the middleware required a session and an admin role.
- Align block explorer docs with permission api docs for access without auth 

Related to matter-labs/zksync-prividium#1019.

## Checklist


aon (approved):

tomimor:
After security review, both /docs will require authentication. Closing PR, keeping auth in place for Prividium Explorer docs.

PR #627, opened 2026-06-25
feat: support custom assets storage and custom base

Romsters:
# What ❔

- support custom custom base
- support serving assets from third-party url
- upload static assets to R2 

## Why ❔

- separate storage for static assets makes old assets accessible 

## How it works

Two independent runtime knobs, carried on `window` by a small inline script in `index.html` (filled by **gomplate** at container start for Docker, or `replaceEnvVariables` at build time for regular builds):

- **`window.__APP_BASE__`** — the app base path. Drives the router (`createWebHistory(appBase)`), `config.js`/favicon, same-origin navigations, and the fallback asset base.
- **`window.__ASSETS_BASE__`** — where hashed assets load from (the CDN origin, optionally a `/<version>/` sub-folder), else the app base.

### Asset base (the tricky part)

Vite 2.9 bakes `base` as a literal string into the bundle, with no runtime hook. To keep **one image** retargetable without rewriting JS on disk (which would desync sourcemaps):

- Every **build** bakes a sentinel base `"/__ASSET_BASE__/"`.
- A small Vite plugin (`runtime-asset-base`, using `magic-string`) rewrites the standalone preload-base constant to `(window.__ASSETS_BASE__||"/")` **at build time**, so the bundle reads the base at runtime and its `.map` stays aligned. A build-time guard throws if any sentinel survives (e.g. a future `import x from "*.svg"`).
- Dynamic `/images` references go through a runtime `resolveAsset()` helper; static `` were converted to it too.
- Only `index.html`'s entry/preload/stylesheet URLs still carry the sentinel; the Docker `CMD` `sed`s those to the chosen base at startup (and injects the CDN origin into the CSP). `config.js` and favicon always stay on the app at `APP_BASE`.

Because the bundle is **byte-identical** across Docker and regular builds, a single Sentry sourcemap upload matches every deployment.

### App base path

`resolveBase()` / `appPrefix` prefix same-origin app paths; `resolveAsset()` handles assets. All base-sensitive spots that bypass the router were updated: the 401 login redirect (`useFetchInstance`), OAuth `redirect_uri` (`useLogin`), MetaMask `blockExplorerUrls` (`useWallet`), OAuth `post_logout_redirect_uri` (`prividium-auth`), logout reload (`WalletInfoModal`), not-found `history.replaceState` (`useNotFound`), the network-switch URL (`NetworkSwitch`), the login redirect query (`App.vue`), post-login navigation (`LoginView`/`AuthCallbackView`), and the header/network logos (`TheHeader` and friends).

## Runtime configuration (container env)

| Var | Default | Purpose |
| --- | --- | --- |
| `APP_BASE` | `/` | Base path the app is served under (e.g. `/explorer/`). |
| `STATIC_ASSETS_URL` | _(empty)_ | CDN/R2 origin to load hashed assets + `/images` from. Empty = serve from the app. |
| `STATIC_ASSETS_VERSIONED` | `true` | When the CDN is set, read assets from a `/<VITE_VERSION>/` sub-folder; `false` = flat. |

> Note: serving under a subpath assumes a **prefix-preserving reverse proxy** in front (nginx is intentionally left mounting at `/`).

## CI / publishing

New `publishStaticAssetsToR2` job (runs **before** the App image is pushed, only when configured): downloads the released `dist.zip` (byte-identical to the image), uploads `assets/` + `images/` to `<upload-folder>/<version>/`, and prunes to the newest **100** versions. The App image push is gated so a real upload failure blocks it, while an unconfigured/skipped upload doesn't.

GitHub configuration:

| Kind | Name | Example / notes |
| --- | --- | --- |
| var | `STATIC_ASSETS_UPLOAD_FOLDER` | `bucket/explorer` — S3 destination; also the on/off switch. Path must match `STATIC_ASSETS_URL`. |
| var | `STATIC_ASSETS_S3_ENDPOINT` | `https://<account>.r2.cloudflarestorage.com` |
| secret | `STATIC_ASSETS_ACCESS_KEY_ID` | R2 S3 API key id |
| secret | `STATIC_ASSETS_SECRET_ACCESS_KEY` | R2 S3 API secret |

## Operator notes

- **CORS**: the bundle's ES-module scripts are fetched cross-origin, so the CDN/R2 bucket **must** return `Access-Control-Allow-Origin` for the app origin(s), or the app won't load.
- The CDN URL must be the bucket's **public** URL (R2.dev or a custom domain), not the `…r2.cloudflarestorage.com` S3 endpoint (that's upload-only).
- `STATIC_ASSETS_UPLOAD_FOLDER` (CI) and `STATIC_ASSETS_URL` (runtime) must point at the same location.

## Testing

- Unit tests, `vue-tsc` typecheck, and ESLint all pass; added `tests/utils/appBase.spec.ts` covering `appBase`/`appPrefix`/`resolveBase`/`resolveAsset` at root, subpath, and CDN/versioned.
- Verified Docker and non-Docker builds produce identical chunk hashes; validated the gomplate + `sed` runtime substitution end-to-end for default / custom-base / CDN(±versioned) modes.
- Prividium auth e2e (`@prividium`) exercises the login/redirect paths touched here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,14 +53,63 @@ jobs:
             VITE_APP_ENVIRONMENT=production \
               npx semantic-release
 
+  # Publish the app's assets + public images to the R2 bucket before the App image is pushed so the files are already in place when it goes live.
+  publishStaticAssetsToR2:
+    name: Publish app static assets to R2
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    needs: createReleaseVersion
+    if: ${{ github.ref == 'refs/heads/main' && needs.createReleaseVersion.outputs.releaseVersion != '' && vars.STATIC_ASSETS_UPLOAD_FOLDER != '' }}
+    steps:
+      - name: Download Dist package
+        uses: dsaltares/fetch-gh-release-asset@master
+        with:
+          version: "tags/v${{ needs.createReleaseVersion.outputs.releaseVersion }}"
+          file: "dist.zip"
+          target: "dist.zip"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Unzip Dist package
+        run: |
+          unzip dist.zip -d packages/app
+
+      - name: Publish to R2
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.STATIC_ASSETS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.STATIC_ASSETS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: auto
+        run: |
+          set -e
+          APP_DIST=packages/app/dist
+          ASSETS_DEST="s3://${{ vars.STATIC_ASSETS_UPLOAD_FOLDER }}"
+          ASSETS_S3_ENDPOINT="${{ vars.STATIC_ASSETS_S3_ENDPOINT }}"
+          VERSION="${{ needs.createReleaseVersion.outputs.releaseVersion }}"
+
+          aws s3 sync "$APP_DIST/assets" "$ASSETS_DEST/$VERSION/assets" --endpoint-url "$ASSETS_S3_ENDPOINT" \
+            --cache-control "public, max-age=31536000, immutable" --exclude "*.map"
+          aws s3 sync "$APP_DIST/images" "$ASSETS_DEST/$VERSION/images" --endpoint-url "$ASSETS_S3_ENDPOINT" \
+            --cache-control "public, max-age=31536000, immutable"
+
+          # Retention: keep only the newest 50 version folders, delete the rest.
+          KEEP=50
+          STALE=$(aws s3 ls "$ASSETS_DEST/" --endpoint-url "$ASSETS_S3_ENDPOINT" \
+            | awk '/ PRE / {print $2}' | sed 's#/$##' | grep -E '^[0-9]' | sort -V | head -n "-$KEEP" || true)
+          for v in $STALE; do
+            echo "Pruning old asset version: $v"
+            aws s3 rm "$ASSETS_DEST/$v/" --recursive --endpoint-url "$ASSETS_S3_ENDPOINT"
+          done
+
   deployBackendToStaging:
     name: Deploy Block Explorer backend to staging
     runs-on: [matterlabs-ci-runner]
     permissions:
       contents: read
       packages: write
-    needs: createReleaseVersion
-    if: ${{ github.ref == 'refs/heads/main' && needs.createReleaseVersion.outputs.releaseVersion != '' }}
+    needs: [createReleaseVersion, publishStaticAssetsToR2]
+    # always() so a skipped R2 job (bucket not configured) doesn't skip this one; but a real R2
+    # failure does block the image push (assets must be live first).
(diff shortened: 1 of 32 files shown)


packages/app/README.md
```diff
@@ -77,11 +77,35 @@ npm run test:e2e
...
+| `STATIC_ASSETS_URL` | _(empty)_ | If set, hashed assets and `/images` load from this origin instead of the app (e.g. `https://static.example.com/explorer`). Empty = serve everything from the app at `APP_BASE`. |
+| `STATIC_ASSETS_VERSIONED` | `true` | When `STATIC_ASSETS_URL` is set, assets are read from a `/<VITE_VERSION>/` sub-folder. Set `false` to read them flat (requires assets uploaded flat — the release pipeline always uploads versioned). |
+
+`config.js` and the favicon are always served by the app at `APP_BASE` (never the CDN). Only the
+content-hashed bundle and `/images` move to `STATIC_ASSETS_URL`.
+
+### Publishing assets to the CDN (CI)
+
+When the release workflow runs, the `publishStaticAssetsToR2` job uploads the released `dist`'s
+`assets/` and `images/` to the bucket under `<version>/`, then prunes to the newest 50 versions. It runs
```
vasyl-ivanchuk:
```suggestion
`assets/` and `images/` to the bucket under `<version>/`, then prunes to the newest 100 versions. It runs
```

packages/app/src/views/AuthCallbackView.vue
```diff
@@ -61,7 +67,7 @@ const redirectToLogin = () => {
 onMounted(async () => {
   try {
     const { redirect } = await handlePrividiumCallback();
-    await router.push(isValidRedirectPath(redirect) ? redirect : "/");
+    window.location.href = isValidRedirectPath(redirect) ? redirect : resolveBase("/");
```
vasyl-ivanchuk, line 70: Since you changed `router.push` to `window.location.href`, does it make sense to tighten the `isValidRedirectPath` check to ensure it points to the same domain?

.github/workflows/release.yml
```diff
@@ -53,14 +53,63 @@ jobs:
...
+  publishStaticAssetsToR2:
+    name: Publish app static assets to R2
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    needs: createReleaseVersion
+    if: ${{ github.ref == 'refs/heads/main' && needs.createReleaseVersion.outputs.releaseVersion != '' && vars.STATIC_ASSETS_UPLOAD_FOLDER != '' }}
+    steps:
+      - name: Download Dist package
+        uses: dsaltares/fetch-gh-release-asset@master
```
vasyl-ivanchuk: nit: `@master` also used in other places, so it's not something introduced in this PR, but we will need to update all GH actions to use specific SHA commit versions.

vasyl-ivanchuk:
@Romsters `packages/app/src/components/contract/ContractInfoTab.vue` has links with a forward slash:
```
<a :href="`/address/${contract.proxyInfo?.implementation.address}#contract`">
```
they are root-relative and won't respect the subpath.

packages/app/src/views/AuthCallbackView.vue
Romsters: It’s still relative path

.github/workflows/release.yml
Romsters: done

Romsters:
> @Romsters `packages/app/src/components/contract/ContractInfoTab.vue` has links with a forward slash:
> 
> ```
> <a :href="`/address/${contract.proxyInfo?.implementation.address}#contract`">
> ```
> 
> they are root-relative and won't respect the subpath.

You are right, I missed that pattern

vasyl-ivanchuk:
LGTM!

vasyl-ivanchuk (dismissed):

vasyl-ivanchuk (approved):

PR #640, opened 2026-07-09
feat: add statusBadgeIconUrl config option for transaction status badge icon

bxpana:
# What ❔

Adds an optional `statusBadgeIconUrl` field to `NetworkConfig`. When set, the transactions table status badge renders the configured image instead of the default ZKsync arrows icon. If the configured image fails to load (404, CDN outage, blocked request), the badge falls back to the default icon via an `@error` handler instead of showing a broken-image glyph.

## Why ❔

Hyperchain operators customizing the explorer (logo, hero banner, theme) currently have no way to replace the ZKsync brand icon inside the transaction status badge. This extends the existing per-network branding options with the same pattern used by `logoUrl` and `heroBannerImageUrl` (optional config field, resolved through `resolveAsset`).

## Checklist

### DIFF
--- a/packages/app/src/components/transactions/Table.vue
+++ b/packages/app/src/components/transactions/Table.vue
@@ -45,11 +45,10 @@
           <template #default>
             {{ te(`transactions.status.${item.status}`) ? t(`transactions.status.${item.status}`) : item.status
             }}<img
-              v-if="currentNetwork.statusBadgeIconUrl && !statusBadgeIconFailed"
+              v-if="currentNetwork.txStatusBadgeIconUrl"
               class="status-badge-icon"
-              :src="resolveAsset(currentNetwork.statusBadgeIconUrl)"
+              :src="resolveAsset(currentNetwork.txStatusBadgeIconUrl)"
               alt=""
-              @error="statusBadgeIconFailed = true"
             /><component :is="item.statusIcon" v-else />
           </template>
         </Badge>
@@ -248,8 +247,6 @@ import { resolveAsset } from "@/utils/appBase";
 import { getContractDisplayName, isContractDeployerAddress, utcStringFromISOString } from "@/utils/helpers";
 
 const { currentNetwork } = useContext();
-
-const statusBadgeIconFailed = ref(false);
 
 const { t, te } = useI18n();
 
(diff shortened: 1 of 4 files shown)


packages/app/src/components/transactions/Table.vue
Romsters: This is the right place to override icon and then component should render whatever is here

packages/app/src/components/transactions/Table.vue
```diff
@@ -44,7 +44,13 @@
         <Badge :color="item.statusColor" :data-testid="$testId.statusBadge">
           <template #default>
             {{ te(`transactions.status.${item.status}`) ? t(`transactions.status.${item.status}`) : item.status
-            }}<component :is="item.statusIcon" />
+            }}<img
+              v-if="currentNetwork.statusBadgeIconUrl && !statusBadgeIconFailed"
+              class="status-badge-icon"
+              :src="resolveAsset(currentNetwork.statusBadgeIconUrl)"
+              alt=""
+              @error="statusBadgeIconFailed = true"
```
Romsters: I'd rather prefer a broken image tbh, at least it's visible that something is wrong.

packages/app/src/configs/index.ts
```diff
@@ -6,6 +6,7 @@ export type NetworkConfig = {
   logoInverseUrl?: string;
   heroBannerImageUrl?: string;
   heroBannerCover?: boolean;
+  statusBadgeIconUrl?: string;
```
Romsters: be specific: txStatusBadgeIconUrl

packages/app/src/components/transactions/Table.vue
bxpana: Fair enough, removed the fallback in 0ffbac77a so a bad URL shows visibly.

packages/app/src/configs/index.ts
bxpana: Renamed to txStatusBadgeIconUrl in 0ffbac77a.

packages/app/src/components/transactions/Table.vue
bxpana: Done in 3899c6437. Moved the override into the computed, statusIcon is now either ZkSyncIcon or a small TxStatusBadgeIcon component that renders the configured image, and the template is back to a single <component :is="item.statusIcon" />.

Romsters (approved):

